PR #370, opened 2024-08-28
Fix stale scrape targets issue

jwilder:
There is a chance that pods can change IPs if the CRI restarts or kills a pod behind the scenes.  This can cause another pod to be assigned an IP that collector is scraping and either scrape the wrong things or continue to fail.

This improves the detection of when a scrape target changes.

### DIFF
--- a/collector/scraper.go
+++ b/collector/scraper.go
@@ -115,7 +115,7 @@ type Scraper struct {
 	cancel context.CancelFunc
 
 	mu      sync.RWMutex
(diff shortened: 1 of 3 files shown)


mkeesey (approved):
Wow the informer interface is obnoxious that there isn't an easy way to just force the update. Saw some examples which took a lot more code to wait for the sync, but I think the `time.Sleep()` calls are better than that.

PR #415, opened 2024-10-16
Design proposal for user specified custom schemas

jessejlt:


docs/designs/schema-etl.md
```diff
@@ -0,0 +1,182 @@
...
+
+```yaml
+apiVersion: adx-mon.azure.com/v1
+kind: Function
+metadata:
+  name: samplefn
+spec:
+  name: SampleFn
+  body: |
+    SampleFn
```
mkeesey, line 153: Should/does the source table in this example need to be different than the Function name, or does it work as-is?

docs/designs/schema-etl.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+## Implementation Details
+
+- Ingestor already manages the creation of Tables as logs flow through the system, so we'll have Ingestor also manage creation of these Functions because the Table must exist prior to the Function that it references.
+- Alerter already handles CRDs, as much as possible we'll share code between these components so that we're not duplicating a bunch of code.
+- When attempting to reconcile each Function's state, we'll make sure the process is idempotent such that if a Function definition already exists that matches all the parameters found in its CRD, we do not attempt to update its state.
+
+## Questions
+- If Ingestor has a reference to a CRD / Function that is later deleted, do we want to delete the associated Function in Kusto?
```
mkeesey, line 182: I would lean to delete if the CRD is no longer present since it is easy to recreate, but open to other thoughts too.

docs/designs/schema-etl.md
jessejlt: They need to be the same if the desired outcome is a View that can be queried in place of a Table (facade)

mkeesey (approved):

PR #420, opened 2024-10-23
CPU Usage Reduction

jwilder:
This has a few changes to reduce CPU.  The first is to use persistent connections with http 1.1 and a gradual drain on the server side.  The second is to make the segment flush interval configurable to reduce CPU usage.

This is before and after of CPU usage for ingeston and collector on a 100 node cluster.  Roughly a 3x decrease in CPU for ingestor and 3-10x for collector.

image

### DIFF
--- a/pkg/http/randomizer.go
+++ b/pkg/http/randomizer.go
@@ -6,7 +6,7 @@ import (
 )
 
 // closeEvery is the chance that a request will be closed.
-const closeEvery = 10000 // every 1000 requests will be closed
+const closeEvery = 10000 // every 10000 requests will be closed
 
 var counter uint64
 
--- a/pkg/http/randomizer.go
+++ b/pkg/http/randomizer.go
@@ -1,0 +1,18 @@
+package http
+
+import (
+	"net/http"
+	"sync/atomic"
+)
+
+// closeEvery is the chance that a request will be closed.
+const closeEvery = 10000 // every 1000 requests will be closed
+
+var counter uint64
+
+func MaybeCloseConnection(w http.ResponseWriter, r *http.Request) {
+	if atomic.AddUint64(&counter, 1)%closeEvery == 0 {
+		r.Close = true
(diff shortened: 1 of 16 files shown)


pkg/http/randomizer.go
```diff
@@ -0,0 +1,18 @@
+package http
+
+import (
+	"net/http"
+	"sync/atomic"
+)
+
+// closeEvery is the chance that a request will be closed.
+const closeEvery = 10000 // every 1000 requests will be closed
```
mkeesey, line 9: nit: every 10000

pkg/http/randomizer.go
```diff
@@ -0,0 +1,18 @@
...
+
+var counter uint64
+
+func MaybeCloseConnection(w http.ResponseWriter, r *http.Request) {
+	if atomic.AddUint64(&counter, 1)%closeEvery == 0 {
```
mkeesey, line 14: I'm not sure this will enforce fairness very well vs per-connection counters, but I think this is fine to move forward.

I'm also a bit worried that under high request rates that we may "skip" closes, but I may misunderstand how AddUint64 works under the hood. My model is:

counter=9999
thread1: add 1
thread2: add 1
thread1: return 10001
thread1: return 10001

mkeesey (approved):

pkg/http/randomizer.go
mkeesey: For per connection ones if we go that way in the future, I've used the `ConnContext` field in `Server` to populate a uint initialized to zero which I have incremented and checked to see if it's time to kick that connection off.

jwilder: Atomic won't allow that.  After thread1 adds, it'll be 10000.  After thread2 add, it's 10002.

pkg/http/randomizer.go
jwilder: fixed

mkeesey (approved):

PR #425, opened 2024-10-30
Set error messages on Function CRDs

jessejlt:
Unify CRD function tasks and views status update
handler to also include an optional error message
to make it easier to diagnose operational failures.


ingestor/adx/tasks.go
```diff
@@ -156,23 +149,39 @@ func (t *SyncFunctionsTask) Run(ctx context.Context) error {
...
+		errMsg := err.Error()
+		if len(errMsg) > 256 {
+			errMsg = errMsg[:256]
+		}
+		fn.Status.Error = errMsg
+	}
+	if err := store.UpdateStatus(ctx, fn); err != nil {
+		logger.Errorf("Failed to update status for function %s.%s: %v", fn.Spec.Database, fn.Name, err)
+	}
+	return nil
```
jwilder, line 186: This function doesn't ever return an error.  I think you can remove the error from the signature.

jessejlt: Done!

jwilder (approved):

PR #426, opened 2024-10-30
Adding journald support

mkeesey:
Major changes in this PR:

1. Renaming the `tail-log` config section to `host-log` to better match `otel-log` and the fact that we can configure file tails and journald within it.
2. Building with CGO. This is necessary to use the native systemd libs to read the journal. Unfortunately this also necessitates a relatively large runtime container to have access to a package manager to install the requisite systemd libs.
3. Modifying our default configs to match the new config names, and to remove some duplicated fields we had been adding to logs.


cmd/collector/config.go
```diff
@@ -365,25 +354,94 @@ func (w *TailLog) Validate() error {
...
+		haveFirstFilter = true
+	}
+
+	for _, parserName := range j.Parsers {
+		if !parser.IsValidParser(parserName) {
+			return fmt.Errorf("parsers %s is not a valid parser", parserName)
+		}
+	}
+
+	return nil
```
jessejlt, line 429: Since `database` and `table` are used as part of the WAL filename, should we do some sort of validation and sanitization to these parameters to prevent any sort of path traversal vulnerabilities?

collector/logs/sources/journal/cursor.go
```diff
@@ -0,0 +1,68 @@
...
+
+func cursorPath(cursorDirectory string, filters []string, database string, table string) string {
+	hasher := xxhash.New()
+	// filters are order dependent, so do not sort before creating the hash.
+	for _, filter := range filters {
+		hasher.Write([]byte(filter))
+	}
+	filterHash := hasher.Sum64()
+	fileName := fmt.Sprintf("journal_%s_%s_%x.cursor", database, table, filterHash)
+	sanitized := strings.ReplaceAll(fileName, string(filepath.Separator), "_")
```
jessejlt, line 26:
```suggestion
	sanitized := filepath.Clean(fileName)
```

Might be a safter way to do the same thing

collector/logs/sources/journal/cursor.go
```diff
@@ -0,0 +1,68 @@
...
+func readCursor(cursorFilePath string) (string, error) {
+	input, err := os.Open(cursorFilePath)
+	if err != nil {
+		return "", fmt.Errorf("journal: failed to open cursor file: %w", err)
+	}
+	defer input.Close()
+
+	var cursor journalcursor
+	if err := json.NewDecoder(input).Decode(&cursor); err != nil {
+		return "", fmt.Errorf("journal: failed to decode cursor file: %w", err)
```
jessejlt, line 58:
```suggestion
		cleanCursor(cursorFilePath)
		return "", fmt.Errorf("journal: failed to decode cursor file: %w", err)
```

Does it make sense to nuke the cursor at this point? Seems that if we can't decode the cursor now, we won't be able to do it later, and will just end up in some sort of infinite failure mode.

jessejlt (approved):

collector/logs/sources/journal/cursor.go
mkeesey: Sadly `Clean` doesn't quite do this despite the name - it just removes duplicate separators and simplifies `..` and `.` bits within the path, but if there are separators in the path it keeps em.

cmd/collector/config.go
mkeesey: This is a good question. I want to see how we're doing this over time.

collector/logs/sources/journal/cursor.go
mkeesey: I *think* we want to keep them in this path at least. I thought about this after your question, and I think a big use case to keep it around is if there is a bad collector release that can't read these correctly - might be nice not to nuke all of these right away.

I plan to add a mechanism for us to clean up old cursor files - if they haven't been written in a while we'll just nuke em.

PR #429, opened 2024-11-04
Make tail source more resilient to individual tailers having errors

mkeesey:
We were previously sharing an errgroup amongst all tailers, which meant that any that returned an error would cause all tailers to stop. Instead, we now have per-tailer errgroups to prevent this from happening.

We also had a mixture of contexts and cancel functions within the source itself and within the tailer/discovery workers along with Open() and Close() functions on those workers. Instead, just use Open() and Close() on the workers.

### DIFF
--- a/collector/logs/sources/tail/tailer.go
+++ b/collector/logs/sources/tail/tailer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"sync"
 	"time"
 
 	"github.com/Azure/adx-mon/collector/logs/engine"
@@ -12,7 +13,6 @@ import (
 	"github.com/Azure/adx-mon/collector/logs/types"
 	"github.com/Azure/adx-mon/pkg/logger"
 	"github.com/tenebris-tech/tail"
-	"golang.org/x/sync/errgroup"
 )
 
 type TailerConfig struct {
@@ -28,7 +28,7 @@ type TailerConfig struct {
 type Tailer struct {
 	tail           *tail.Tail
 	shutdown       context.CancelFunc
-	errgroup       *errgroup.Group
+	wg             sync.WaitGroup
 	database       string
 	table          string
 	logTypeParser  sourceparse.LogTypeParser
@@ -37,8 +37,7 @@ type Tailer struct {
 }
 
 func StartTailing(config TailerConfig) (*Tailer, error) {
-	group, ctx := errgroup.WithContext(context.Background())
-	ctx, shutdown := context.WithCancel(ctx)
+	ctx, shutdown := context.WithCancel(context.Background())
 
 	batchQueue := make(chan *types.Log, 512)
 	outputQueue := make(chan *types.LogBatch, 1)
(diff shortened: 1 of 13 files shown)


collector/logs/sources/tail/tailer.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		tailConfig.Location = &tail.SeekInfo{
+			Offset:         position,
+			Whence:         io.SeekStart,
+			FileIdentifier: fileId,
+		}
+	}
+
+	tailFile, err := tail.TailFile(config.Target.FilePath, tailConfig)
+	if err != nil {
```
jessejlt, line 60: Just a comment and nothing to block the current PR. I wonder if we should consider taking inventory of the type of failure modes we can encounter at this point and try to distinguish between permanent and transitive, such that if the error is transitive we can attempt a retry. My concern is that, if my understanding of the code is correct, we essentially abandon tailing a file that returns an error for any reason, which means we in effect _lose_ those logs, especially if a rotation occurs.

collector/logs/sources/tail/tailer.go
```diff
@@ -0,0 +1,201 @@
...
+		shutdown:       shutdown,
+		errgroup:       group,
+		database:       config.Target.Database,
+		table:          config.Target.Table,
+		logTypeParser:  sourceparse.GetLogTypeParser(config.Target.LogType),
+		logLineParsers: parsers,
+		resources:      attributes,
+	}
+	group.Go(func() error {
+		return readLines(ctx, tailer, config.UpdateChan, batchQueue)
```
jessejlt, line 87: Along the same thought process as the previous comment, it seems that returning an error into the error group due to what could easily be a transitive failure, causing us to abandon tailing this file, is something we're going to want to take a closer look at in the future. I think we should error on the side of robustness / permissive error handling when tailing these files such that we integrate a reasonable retry strategy into any error that is transitive in nature; or maybe even treat all errors that are transitive except for a select few that we know are permanent, like `os.ErrNotExist`.

collector/logs/sources/tail/tailer.go
```diff
@@ -0,0 +1,201 @@
...
+// Wait waits for the tailer to finish processing.
+func (t *Tailer) Wait() {
+	t.errgroup.Wait()
+}
+
+func readLines(ctx context.Context, tailer *Tailer, updateChannel <-chan FileTailTarget, outputQueue chan<- *types.Log) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
```
jessejlt, line 125: Should we return `ctx.Err()` so that the caller, in this case an `errgroup` correctly terminates all other members of that group?

collector/logs/sources/tail/tailer.go
```diff
@@ -0,0 +1,201 @@
...
+
+			// Write after parsing to ensure these values are always set to values we need for acking.
+			log.Attributes[cursor_position] = position
+			log.Attributes[cursor_file_id] = currentFileId
+			log.Attributes[cursor_file_name] = tailer.tail.Filename
+
+			outputQueue <- log
+		}
+	}
+}
```
jessejlt, line 201: I see this function is very permissive in its error handling, so my previous comments are maybe not warranted, though I suppose the danger is that a future maintainer doesn't quite follow the same implicit permissive error handling in this function. I'm not really sure how to address making explicit the principal that this function should not abate unless the intent is to abandon the file completely.

jessejlt (approved):

mkeesey:
I cleaned up how we handle errors in these cases by moving away from errgroup (which allows errors in the callers) to plain goroutines. In practice I had eliminated all non-nil error returns after the first commit which was the underlying bug in this case.

I think this makes it much clearer that the pattern should be not returning errors from handlers, but instead more gracefully handling these cases.

PR #436, opened 2024-11-12
Don't block appends when segments are highly contended

jwilder:
Under higher write load, the segment write lock can cause
long delays for a goroutine to acquire it for a write and append
operation.  This causes ingestor to report "slow request" at times
because multiple locks are getting blocked for while and the request
goes slower.

For collector shipping segments, it generally has many segements
on disk that it needs to send.  So instead of blocking when a segment
is locked, we now just return a locked error and let collector move
to the next segment.  It will retry the prior segment again in a few
seconds.  This allows for better and reduces the slow requests warnings
quite a bit.

### DIFF
--- a/cmd/ingestor/main.go
+++ b/cmd/ingestor/main.go
@@ -408,6 +408,14 @@ func realMain(ctx *cli.Context) error {
 	srv.ErrorLog = newLogger()
 
 	go func() {
+		// Under high connection load and when the server is doing a lot of IO, this
+		// can cause the server to be unresponsive.  This pins the accept goroutine
+		// to a single CPU to reduce context switching and improve performance.
+		runtime.LockOSThread()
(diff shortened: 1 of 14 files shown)


cmd/ingestor/main.go
```diff
@@ -408,6 +409,14 @@ func realMain(ctx *cli.Context) error {
 	srv.ErrorLog = newLogger()
 
 	go func() {
+		// Under high connection load and when the server is doing a lot of IO, this
+		// can cause the server to be unresponsive.  This is pins the accept goroutine
```
jessejlt, line 413:
```suggestion
		// can cause the server to be unresponsive.  This pins the accept goroutine
```

jessejlt (approved):

mkeesey (approved):

jessejlt (approved):

PR #446, opened 2024-11-19
Update quickstart docs with more examples, add config doc

mkeesey:
The quickstart docs now include more examples of how to configure pods to have logs scraped. Additionally, there are some example KQL queries.

We now generate a config doc based on the config struct, which allows us to keep the doc and the config in sync and reuse the comments associated with the toml elements within source. To better support this, I've configured many of the global config options as `omitempty` in the struct to more easily control the serialized outputs we write to the markdown file.


jwilder (approved):

mkeesey:

PR #456, opened 2024-11-25
setup.sh grants AllDatabasesAdmin to support cluster-wide functions

mkeesey:


mkeesey:
This is probably not needed.

PR #457, opened 2024-11-26
Only use systemd libs if we are building with CGo

mkeesey:
If statically linking (or testing with static linking), do not require CGo if we don't have it enabled.


jwilder (approved):

mkeesey:

PR #458, opened 2024-11-26
Add e2e tests

jessejlt:
Integration test for the entire pipeline


pkg/testutils/integration_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	ingestorContainer, err := ingestor.Run(ctx, ingestor.WithCluster(ctx, k3sContainer))
+	testcontainers.CleanupContainer(t, ingestorContainer)
+	require.NoError(t, err)
+
+	collectorContainer, err := collector.Run(ctx, collector.WithCluster(ctx, k3sContainer))
+	testcontainers.CleanupContainer(t, collectorContainer)
+	require.NoError(t, err)
+
+	t.Run("Logs", func(t *testing.T) {
```
mkeesey, line 55: Can we set the Logs and Metrics blocks to run with `t.Parallel()` so they can wait in parallel?

pkg/testutils/kustainer/kustainer.go
```diff
@@ -260,3 +261,14 @@ func (c *KustainerContainer) CreateDatabase(ctx context.Context, dbName string)
 	}
 	return nil
 }
+
+type threadsafeBuffer struct {
+	sync.Mutex
+	buffer bytes.Buffer
+}
+
+func (b *threadsafeBuffer) Write(p []byte) (n int, err error) {
```
mkeesey, line 270: Do we need a wrapper to read the buffer? I think we probably need to have something that wraps reads with a lock operation too.

pkg/testutils/integration_test.go
```diff
@@ -0,0 +1,106 @@
...
+	t.Logf("Kubeconfig: %s", kubeconfig)
+	t.Logf("Kustainer: %s", kustoContainer.ConnectionUrl())
+
+	t.Run("Create databases", func(t *testing.T) {
+		for _, dbName := range []string{"Metrics", "Logs"} {
+			require.NoError(t, kustoContainer.CreateDatabase(ctx, dbName))
+		}
+	})
+
+	ingestorContainer, err := ingestor.Run(ctx, ingestor.WithCluster(ctx, k3sContainer))
```
mkeesey, line 47: For a follow on PR, I think it would be nice to have a way to configure the Kusto batch policies so we can set them to something like 30s for logs (instead of the 5m we hard code them to) which hopefully speeds this test up a lot.

pkg/testutils/integration_test.go
jessejlt: Everything that can be ran in parallel is now doing so. I decided to use go routines to execute blocks in parallel instead of `t.Parallel()` due to difficulties with the way Go executes parallel tests.

pkg/testutils/kustainer/kustainer.go
jessejlt: I don't think so, but I added one.

pkg/testutils/integration_test.go
jessejlt: Added this functionality, which was a bit more complicated than I would have thought due to the time formatting required by Kusto.

mkeesey (approved):

PR #464, opened 2024-12-03
Implement cache management for Kusto functions and add test for CRD deletions

jessejlt:
When a Function CRD has been seen and cached by Ingestor, if that CRD is later deleted from k8s, we detect the deletion and remove the corresponding function from Kusto.

### DIFF
--- a/ingestor/adx/tasks.go
+++ b/ingestor/adx/tasks.go
@@ -5,6 +5,7 @@ import (
 	ERRS "errors"
 	"fmt"
 	"io"
+	"slices"
 	"sync"
 	"time"
 
@@ -125,18 +126,38 @@ func NewSyncFunctionsTask(store FunctionStore, kustoCli StatementExecutor) *Sync
 	}
 }
 
+func (t *SyncFunctionsTask) cacheKey(fn *v1.Function) string {
+	return fn.Namespace + fn.Name
+}
+
 func (t *SyncFunctionsTask) Run(ctx context.Context) error {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 
 	functions := t.store.Functions()
+	for cacheKey, cacheFn := range t.cache {
+		if !slices.ContainsFunc(functions, func(f *v1.Function) bool {
+			if f.Spec.Database != t.kustoCli.Database() {
+				return false
+			}
+			return t.cacheKey(f) == cacheKey
+		}) {
+			// Function no longer exists, remove from cache and Kusto
+			stmt := kql.New(".drop function ").AddUnsafe(cacheFn.Name)
+			if _, err := t.kustoCli.Mgmt(ctx, stmt); err != nil {
+				logger.Warnf("Failed to drop function %s: %v", cacheFn.Name, err)
+			}
+			delete(t.cache, cacheKey)
+		}
+	}
+
 	for _, function := range functions {
 
 		if function.Spec.Database != t.kustoCli.Database() {
 			continue
 		}
 
-		cacheKey := function.Namespace + function.Name
+		cacheKey := t.cacheKey(function)
 		if fn, ok := t.cache[cacheKey]; ok {
 			if function.Generation != fn.Generation {
 				// invalidate our cache
--- a/ingestor/adx/tasks_test.go
+++ b/ingestor/adx/tasks_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"slices"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


ingestor/adx/tasks.go
```diff
@@ -125,18 +126,38 @@ func NewSyncFunctionsTask(store FunctionStore, kustoCli StatementExecutor) *Sync
...
+	return fn.Namespace + fn.Name
+}
+
 func (t *SyncFunctionsTask) Run(ctx context.Context) error {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 
 	functions := t.store.Functions()
+	for cacheKey, cacheFn := range t.cache {
+		if !slices.ContainsFunc(functions, func(f *v1.Function) bool {
```
mkeesey, line 139: I may misunderstand how everything hooks together, but I think this approach could be pretty dangerous. The population of the cache is asynchronous from this and separate from this, so I could see a couple possible situations:

1. The cache hydration process takes a while (api server is slow, the cache poller starts up slower, etc), so we start with an empty cache and start nuking functions.
2. We don't have a way to communicate from the cache that it has errors populating itself, so if we can't even query for the CRDs we don't have a way in this task to know we shouldn't try to clean anything.

jessejlt: We only delete if a CRD is no longer being returned by the API server where it once was, so an initial empty cache won't cause deletions. In fact, we don't even know what to delete unless it's first represented in the cache. 

Regarding the second point, we only populate the cache if we're able to successfully retrieve CRDs from the API server, so transitive errors have no impact on our internal state representation. 

The flow is as follows.

1. [List CRDs](https://github.com/Azure/adx-mon/blob/1cbe0a724e0f1f189e2891adc1b0fcd7b23f9e0b/pkg/crd/crd.go#L67-L75)

```go
list, err := c.List(ctx)
if err != nil {
	logger.Errorf("Failed to list objects: %s", err.Error())
	continue
}
if err := c.opts.Store.Receive(ctx, list); err != nil {
	logger.Errorf("Failed to store objects: %s", err.Error())
	continue
}
```

Here we see that if an error occurs during the CRD listing process (this is where the API server is queried), we simply wait until the next poll interval or notification and log the error. Notably, we only invoke the [receiver](https://github.com/Azure/adx-mon/blob/1cbe0a724e0f1f189e2891adc1b0fcd7b23f9e0b/ingestor/storage/kql_functions.go#L89) upon successful querying of the API server / listing of CRDs. 

2. Upon successful listing of CRDs, we invoke our receiver.

```go
func (f *Functions) Receive(ctx context.Context, list client.ObjectList) error {
	items, ok := list.(*v1.FunctionList)
	if !ok {
		return fmt.Errorf("expected *v1.FunctionList, got %T", list)
	}
	if items == nil || len(items.Items) == 0 {
		return nil
	}

	var (
		views     = &v1.FunctionList{}
		functions = &v1.FunctionList{}
		unique    = make(map[string]struct{})
	)
	for _, function := range items.Items {
		// TODO (jesthom): If a database isn't specified, should we just consider installing
		// the function in all tracked databases? This would be useful for adx-mon's
		// own functions, for example.
		if function.Spec.Database == "" {
			logger.Errorf("Function %s has no database", function.Name)
			continue
		}
		// TODO (jesthom): Once the parser is in place we can perform more validation.
		if _, ok := unique[function.Spec.Database+function.Name]; ok {
			logger.Errorf("Function %s is a duplicate", function.Name)
			continue
		}
		unique[function.Spec.Database+function.Name] = struct{}{}
		// TODO (jesthom): Once the parser is in place separate out the Views from the Functions.
		functions.Items = append(functions.Items, *function.DeepCopy())
	}

	f.mu.Lock()
	f.views = views
	f.functions = functions
	f.mu.Unlock()

	return nil
}
```

All we're doing here is collating our CRDs into KQL Functions and Views. This process is more of a placeholder until we can properly parse our KQL though we have a slight amount of validation that takes place so as to not pass along invalid CRDs. 

3. [Reconcile](https://github.com/Azure/adx-mon/blob/main/ingestor/adx/tasks.go#L128) Kusto's state using our CRDs as the desired state. 

An empty cache either means there are no CRDs in k8s or we haven't yet populated our cache, there is therefore no action necessary. We can't actually delete a function in Kusto unless that function has first been populated in our cache. 

So we can think of three processes in the flow:
1. Generic [CRD Listing](https://github.com/Azure/adx-mon/blob/main/pkg/crd/crd.go) that knows how to interact with the API server to retrieve a specific CRD Kind. 
2.  [KQL Processing](https://github.com/Azure/adx-mon/blob/main/ingestor/storage/kql_functions.go) where we perform validation and collation of the KQL represented in the CRD. 
3. [Reconciliation](https://github.com/Azure/adx-mon/blob/main/ingestor/adx/tasks.go#L128) of the Kusto state using CRDs as our desired state.

jessejlt: @mkeesey after considering the aforementioned flow, I think what we might want to do for deletion reconciliation is to trigger them not by the absence of a CRD that once existed, but instead by responding to a CRD deletion even directly. I'm going to close this PR and rework the flow a little bit so we can directly respond to deletion events.

PR #466, opened 2024-12-06
Replace our KQL CRD polling mechanism with a controller

jessejlt:
Utilize _kubebuilder's_ generated controller to reconcile Kusto's state using KQL CRDs as the desired state. Our current polling mechanism doesn't accurately capture deletions like a controller will. 

Right now this is dead code and just posted for feedback. If we like this approach, the PR will be modified to delete all the existing KQL CRD handling code and wire-up this controller to Ingestor's main via a [controller-manager](https://book.kubebuilder.io/cronjob-tutorial/empty-main)


mkeesey (approved):
I think this approach is ok. For leader election (which is not hooked up yet), I think we should use the elector.IsLeader() method that we use for several other things in Ingestor. This will then support using a non-k8s function reconciler that just looks at files on disk, for example, so we don't have to use k8s.

PR #494, opened 2024-12-21
try and dump a junit xml 

paulgmiller:
This lets jenkins and azrue devops report things when we lint


jwilder:
Not quite sure what this is for so closing this out.

PR #499, opened 2025-01-06
Give more time to nightlies

jessejlt:
Running our nightlies with race conditions enabled in the integration tests is taking a very long time. This change gives more time to the tests to prevent timing out.

Passing test https://github.com/Azure/adx-mon/actions/runs/12637897253


pkg/testutils/alerter/alerter.go
```diff
@@ -32,7 +32,6 @@ func Run(ctx context.Context, opts ...testcontainers.ContainerCustomizer) (*Aler
 	}
 
 	req := testcontainers.ContainerRequest{
-		Name: "alerter" + testcontainers.SessionID(),
```
jessejlt, line 35: These session-ids were creating instability. Session-ids are suppose to be stable across test packages; however, they were causing issues where the session-id would change after a container is built, thereby causing the test to fail due to an inability to find the container by name. The session-ids weren't in-use anyways.

jwilder (approved):

PR #500, opened 2025-01-06
Log scrape errors at warn level

jwilder:


mkeesey (approved):

collector/scraper.go
```diff
@@ -261,11 +262,11 @@ func (s *Scraper) scrapeTargets(ctx context.Context) {
 			wr = s.flushBatchIfNecessary(ctx, wr)
 		}
 		if err := iter.Err(); err != nil {
-			logger.Errorf("Failed to scrape %s: %s", target.Addr, err.Error())
+			logger.Warnf("Failed to scrape %s/%s/%s at %s: %s", target.Namespace, target.Pod, target.Container, target.Addr, err.Error())
```
vpatelsj, line 265: the log line here clashes with line220. May be a slight reword will help clarify where the error is coming from.

vpatelsj (changes requested):

collector/scraper.go
jwilder: Updated

PR #510, opened 2025-01-15
Create OTLP/HTTP metric exporter

mkeesey:
Sometimes it is helpful to create a fork of metrics flowing through collector to forward a subset onto destinations other than Ingestor and Azure Data Explorer. This PR adds a mechanism called an "exporter" that allows metric collectors to export metrics they have collected to an OTLP/HTTP endpoint in parallel with the metrics they are already sending to Ingestor.


cmd/collector/config/config.go
```diff
@@ -107,6 +107,7 @@ type Config struct {
 	OtelLog               *OtelLog                 `toml:"otel-log,omitempty" comment:"Defines an OpenTelemetry log endpoint. Accepts OTLP/HTTP."`
 	OtelMetric            []*OtelMetric            `toml:"otel-metric,omitempty" comment:"Defines an OpenTelemetry metric endpoint. Accepts OTLP/HTTP and/or OTLP/gRPC."`
 	HostLog               []*HostLog               `toml:"host-log,omitempty" comment:"Defines a host log scraper."`
+	Exporters             *Exporters               `toml:"exporters,omitempty" comment:"Optional configuration for exporting telemetry outside of adx-mon."`
```
jwilder, line 110: Exporter is also the name that prometheus uses for components that expose metrics.  I wonder if this would be confusing.  Only other name that comes to mind is "Forwarders."

jwilder: Any OTEL uses exporter for what this is doing as well, so maybe a moot point.

collector/export/metric_otlp.go
```diff
@@ -0,0 +1,299 @@
...
+			},
+		}
+
+		var nameBytes []byte
+		attributes := make([]*commonv1.KeyValue, 0, len(ts.Labels))
+		for _, element := range ts.Labels {
+			if bytes.Equal(element.Name, nameLabel) {
+				nameBytes = element.Value
+			} else if !bytes.HasPrefix(element.Name, []byte("adxmon_")) {
+				attribute := c.stringKVPool.Get(0).(*commonv1.KeyValue)
```
jwilder, line 215: Does attribute need to be reset after getting it from the pool?  Wondering if there is any chance for prior used data to still be present.

collector/export/metric_otlp.go
```diff
@@ -0,0 +1,299 @@
...
+				// only accept stringval here to avoid allocations
+				stringval := attribute.Value.Value.(*commonv1.AnyValue_StringValue)
+				stringval.StringValue = string(element.Value)
+				attributes = append(attributes, attribute)
+			}
+		}
+		metric.Name = string(nameBytes)
+
+		for _, sample := range ts.Samples {
+			datapoint := c.datapointPool.Get(0).(*metricsv1.NumberDataPoint)
```
jwilder, line 226: `datapoint.Reset()`?

jwilder (approved):

collector/export/metric_otlp.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+
+		scopeMetrics.Metrics = append(scopeMetrics.Metrics, metric)
+	}
+
+	serialized, err := proto.Marshal(exportRequest)
+
+	for _, scopeMetric := range scopeMetrics.Metrics {
+		// shared attribute with all datapoints under metric
+		datapoint := scopeMetric.GetGauge().GetDataPoints()[0]
```
jessejlt, line 245: Is it possible for GetDataPoints to return a nil slice that would cause a panic by indexing into it? Should we check slice length?

ingestor/metrics/handler.go
```diff
@@ -163,3 +160,14 @@ func (s *Handler) HandleReceive(w http.ResponseWriter, r *http.Request) {
 	m.WithLabelValues(strconv.Itoa(http.StatusAccepted)).Inc()
 	w.WriteHeader(http.StatusAccepted)
 }
+
+func (s *Handler) Write(ctx context.Context, req *prompb.WriteRequest) error {
+	g, gCtx := errgroup.WithContext(ctx)
+	for _, writer := range s.requestWriters {
+		g.Go(func() error {
```
jessejlt, line 167: I think the loop capture thing is fixed in Go 1.22 but worth verifying

```suggestion
		writer := writer
		g.Go(func() error {
```

jessejlt (approved):

cmd/collector/config/config.go
mkeesey: Going to stick with the otel convention for now

PR #515, opened 2025-01-16
Document the use of Functions CRD

jessejlt:


docs/concepts.md
```diff
@@ -44,6 +44,30 @@ example, the `pod` label may be common to all metrics and can be pulled up to a
...
+      | project msg = tostring(Body.msg),
+          lvl = tostring(Body.lvl),
+          ts = todatetime(Body.ts),
+          namespace = tostring(Body.namespace),
+          container = tostring(Body.container),
+          pod = tostring(Body.pod),
+          host = tostring(Body.host) 
+    }
+  database: Logs
+```
```
jwilder, line 69: Maybe add a sample query that uses this view for reference?

jwilder (approved):

docs/concepts.md
jwilder: Also might be helpful to point out the view/table precedence in kusto when they are named the same thing.

jessejlt: Done!

PR #516, opened 2025-01-16
Add docs for exporter configuration

mkeesey:


docs/config.md
```diff
@@ -93,6 +93,38 @@ lift-attributes = [
   cluster = 'cluster1'
   geo = 'eu'
 
+# Optional configuration for exporting telemetry outside of adx-mon.
+# Exporters are declared here and referenced by name in each collection source.
```
jwilder, line 97: Maybe add that exporters send a copy of the matched metrics to the destination and that the metric is still send to kusto as well?

jwilder: Might also be useful to explain the flow and how/when the various keep/drop rules are applied with the various sources.  For example, the exporters apply these rules after all the source level rules are applied.

jwilder (approved):

docs/config.md
```diff
@@ -93,6 +93,38 @@ lift-attributes = [
...
+      '^kube_pod_ips$'
+    ]
+    # Regexes of metrics to keep if they have the given label and value.
+    keep-metrics-with-label-value = []
+
+    # Key/value pairs of labels to add to all metrics.
+    [exporters.otlp-metric-export.add-labels]
+      forwarded_to = 'otlp'
+
+    # Labels to drop if they match a metrics regex in the format <metrics regex>=<label name>.
```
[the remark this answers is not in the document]
mkeesey: Fixed in the several places where I had this wrong.

docs/config.md
mkeesey: Updated this comment a bit, and expanded a lot more on it in the Exporters section.

jwilder (approved):

PR #522, opened 2025-01-22
Add "Publish container images" GH workflow

tanayk2610:
This change adds a workflow which:
- Builds container images for the alerter, collector, and ingestor.
- Tags them latest if triggered by a push to main branch or with the release tag if being triggered by a tag push to the repository.
- Pushes the images to the GitHub Container Registry.

Also adds a RELEASE.md file explaining the release process.

### DIFF
--- a/.github/workflows/build-pr.yaml
+++ b/.github/workflows/build-pr.yaml
@@ -22,7 +22,7 @@ jobs:
       with:
         go-version: 1.22
 
-    - name: Install depdenencies
+    - name: Install dependencies
       run: sudo apt install -y libsystemd-dev # for collector journal input
 
     - name: Build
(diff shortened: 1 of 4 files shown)


jwilder:
Not sure this is quite what we want.  We don't want a new tag/release for every merge to main.  main will have latest changes and we'll need to decide when to cut a new release with a tag.

jwilder (commented):
What would a sample release tag look like?  Say we want to release a v0.1.0 version.  What tag would we use and push?

jwilder:
Could you start a RELEASE.md doc that documents the process as well?

tanayk2610:
> What would a sample release tag look like? Say we want to release a v0.1.0 version. What tag would we use and push?

The image tags would be the same as the release tag, i.e. `v0.1.0`

jwilder:
Looks great!

jwilder (approved):

PR #524, opened 2025-01-23
Add error exception for CRD type not found

jessejlt:
By upgrading our dependencies the way that CRD kind not found has changed to include error type NoResourceMatchError. This PR updates the error handling expectations to include NoResourceMatchError when interacting with CRDs, mostly from a testing perspective.


jwilder (approved):

jwilder:
What dependency changed the behavior?

jessejlt:
> What dependency changed the behavior?

My guess is something in k8s.io though I haven't spent the time to dig into it too much other than to see the change in behavior in the error being returned.

PR #530, opened 2025-01-24
fix(*): use updated label syntax in dockerfile

matucker-msft:
fixes:
```
LegacyKeyValueFormat: "LABEL key=value" should be used instead of legacy "LABEL key value" format (line 17)
```


tanayk2610:
Can you update `alerter` and `ingestor Dockerfiles` as well?

matucker-msft:
of course, sorry, had them in my other nit branch (rofl)

tanayk2610 (approved):

PR #536, opened 2025-01-25
Add version information to binaries

tanayk2610:


.github/workflows/publish-container-images.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
           images: ${{ matrix.image }}
           # Use latest tag on default branch (main), otherwise use the tag
           tags: |
-            type=raw, value=latest, enable={{is_default_branch}}
+            # type=raw, value=latest, enable={{is_default_branch}}
```
jwilder, line 54: Intentional to comment this out?

build/images/Dockerfile.alerter
```diff
@@ -5,7 +5,14 @@ RUN tdnf install -y golang ca-certificates
 ADD . /code
 WORKDIR /code
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/alerter ./cmd/alerter
+ARG VERSION GIT_COMMIT BUILD_TIME
```
jwilder, line 8: These args are not in the other two Dockerfiles.  Are they needed?

pkg/version/version.go
```diff
@@ -0,0 +1,13 @@
...
+import "fmt"
+
+var (
+	GitCommit string
+	BuildTime string
+	Version   string
+)
+
+func String() string {
+	return fmt.Sprintf("%s, Commit:%s, Build-time:%s", Version, GitCommit, BuildTime)
```
jwilder, line 12: I think you could use os.ARGS[0] to get the binary name and then not need to add alerter/collector/ingestor in the other spots.

jwilder (changes requested):

cmd/ingestor/main.go
```diff
@@ -93,6 +96,8 @@ func main() {
 }
 
 func realMain(ctx *cli.Context) error {
+	logger.Infof("%s version:%s", os.Args[0], version.String())
```
jwilder, line 99: Would it make sense to move this fmt string into `version.String()` too?

Then all these statements could just be `logger.Infof(version.String())`

jwilder (approved):

cmd/ingestor/main.go
jwilder: Sounds good

PR #538, opened 2025-01-25
Add WalkLabels to RequestTransformer to prevent races

mkeesey:
Downstream of a given metrics collector, multiple downstream RemoteWrite implementations may wish to futher filter the metrics they intend to send. Using TransformWriteRequest or TransformTimeSeries directly is problematic in this case because it modifies the writerequest object in-place. This would interfere with other writers who are concurrently working with the same wr.

Instead, we now have a new function that requires a callback that is called for each label value that has passed the filters within the RequestTransformer. Downstream users copy the values they need lazily instead of requiring users to do full deep copies each time they wish to filter down their set of metrics. In combination with the existing ShouldDropMetric implemenation, this allows for this concurrent filtering without interfering with one another.

### DIFF
--- a/transform/transformer.go
+++ b/transform/transformer.go
@@ -152,6 +152,52 @@ func (f *RequestTransformer) TransformTimeSeries(v *prompb.TimeSeries) *prompb.T
 	return v
 }
 
+// WalkLabels operates similarly to TransformTimeSeries, but instead of modifying the TimeSeries, it calls the callback with the key and value
+// This is safe to call in parallel if the name and value bytes are not modified by the callback.
+func (f *RequestTransformer) WalkLabels(v *prompb.TimeSeries, callback func(name []byte, value []byte)) {
+	f.init()
+
+	var skipLabel bool
+
+	for _, l := range v.Labels {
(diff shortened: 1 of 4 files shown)


transform/transformer.go
```diff
@@ -152,6 +152,52 @@ func (f *RequestTransformer) TransformTimeSeries(v *prompb.TimeSeries) *prompb.T
 	return v
 }
 
+// FilterLabels operates similarly to TransformTimeSeries, but instead of modifying the TimeSeries, it calls the callback with the key and value
+// This is safe to call in parallel if the name and value bytes are not modified by the callback.
+func (f *RequestTransformer) FilterLabels(v *prompb.TimeSeries, callback func(name []byte, value []byte)) {
```
jwilder, line 157: Is `FilterLabels` an accurate name here?  It looks like the callback doesn't return a bool and the labels are not mutated so I wonder if something like `WalkLabels` (like https://pkg.go.dev/path/filepath#WalkDir) might be more appropriate.

jwilder (approved):

PR #539, opened 2025-01-26
Convert partmap to use generic values

jwilder:
This prevents some type safety issues where the values could be cast to the wrong type incorrectly.


jessejlt (approved):
Great tests!

PR #543, opened 2025-01-27
Rework batch scheduling to track segment paths vs prefix

jwilder:
This reworks the scheduler and batch to track specific segment files
instead of prefixes.  This should allow for better throughput of
big files.

In the prior implementation, when there are multiple large segments
that exceed the minUploadSize for the same prefix, we would create
one batch for that file and skip the rest.  The prefix would then
be tagged as part of a batch so subsequent batching rounds would
skip over the other files.  This causes more latency for each of the
files because it takes longer to upload a single larget segment and
we don't queue up additional ones in the mean time.


jessejlt (approved):
This should really help, well done!

PR #544, opened 2025-01-27
[integration test] Add a test to ensure components can be upgraded

jessejlt:
This PR modifies our existing integration test to first install the latest released version of our components. We then build our components and upgrade the previous version to the new, ensuring a clean upgrade path. Existing validation takes place to ensure we don't break anything as part of the version change process.


pkg/testutils/collector/collector.go
```diff
@@ -163,23 +180,66 @@ func WithCluster(ctx context.Context, k *k3s.K3sContainer) testcontainers.Custom
...
+						_, err = clientset.CoreV1().ConfigMaps(collectorConfigMap.Namespace).Update(ctx, collectorConfigMap, metav1.UpdateOptions{})
+						if err != nil {
+							return fmt.Errorf("failed to update configmap: %w", err)
+						}
+					}
+
+					existingDaemonSet, err := clientset.AppsV1().DaemonSets(daemonset.Namespace).Get(ctx, daemonset.Name, metav1.GetOptions{})
 					if err != nil {
-						return fmt.Errorf("failed to create configmap: %w", err)
+						if apierrors.IsNotFound(err) {
```
jwilder, line 203: This is a lot of nested if/then/else.  Is there a way we could simplify this and reduce the nesting?

pkg/testutils/integration_test.go
```diff
@@ -85,14 +85,24 @@ func StartCluster(ctx context.Context, t *testing.T) (kustoUrl string, k3sContai
 		}
 	})
 
-	t.Run("Build and install Ingestor and Collector", func(tt *testing.T) {
-		ingestorContainer, err := ingestor.Run(ctx, ingestor.WithCluster(ctx, k3sContainer))
+	t.Run("Install Ingestor and Collector", func(tt *testing.T) {
+		ingestorContainer, err := ingestor.Run(ctx, "ghcr.io/azure/adx-mon/ingestor:latest", ingestor.WithCluster(ctx, k3sContainer))
```
jwilder, line 89: Maybe we should have a logical `release` type of tag that we move to the latest official release vs depending on `latest`?

jwilder (approved):

pkg/testutils/collector/collector.go
jessejlt: I'm glad you made this comment, it forced me to look at alternatives and I think the Patch option really cleans things up. I'm glad you made this comment, it forced me to look at alternatives and I think the Patch option really cleans things up.

pkg/testutils/integration_test.go
jessejlt: Looks like we're still publishing to `latest` https://github.com/Azure/adx-mon/blob/main/.github/workflows/publish-container-images.yaml but I agree that it would be nice to have some sort of deterministic tag such that our nightly tests could target the previous night's release. I'll add it as an issue to track.

PR #545, opened 2025-01-27
Add common WriteRequest function to correctly write to multiple dests

mkeesey:
The errgroup impl that was previously used was problematic where one remote that was failing could prevent the other destinations from succeeding because the context from errgroup would be cancelled. We would also only capture the first error returned. We now allow all remote writers to complete and capture all of the errors returned into a composite error.

### DIFF
--- a/collector/otlp/metrics.go
+++ b/collector/otlp/metrics.go
@@ -15,7 +15,6 @@ import (
 	"github.com/Azure/adx-mon/pkg/prompb"
 	"github.com/Azure/adx-mon/pkg/remote"
 	"github.com/Azure/adx-mon/transform"
-	"golang.org/x/sync/errgroup"
 )
 
 var (
@@ -87,7 +86,7 @@ type OltpMetricWriterOpts struct {
 	MaxBatchSize int
 
 	Clients       []remote.RemoteWriteClient
-	HealthChecker interface { IsHealthy() bool }
+	HealthChecker interface{ IsHealthy() bool }
 }
 
 type OltpMetricWriter struct {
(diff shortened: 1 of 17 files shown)


jessejlt:
> one remote that was failing could prevent the other destinations from succeeding because the context from errgroup would be cancelled

Good catch!

collector/otlp/metrics.go
```diff
@@ -218,14 +217,8 @@ func (t *OltpMetricWriter) sendBatch(ctx context.Context, wr *prompb.WriteReques
...
-	g, gCtx := errgroup.WithContext(ctx)
-	for _, remoteClient := range t.remoteClients {
-		remoteClient := remoteClient
-		g.Go(func() error {
-			return remoteClient.Write(gCtx, wr)
-		})
-	}
-	if err := g.Wait(); err != nil {
+	err := remote.WriteRequest(ctx, t.remoteClients, wr)
+	if err != nil {
```
jessejlt, line 221: Nice! Much cleaner.

pkg/remote/types.go
```diff
@@ -29,3 +31,26 @@ func (n *nopCloser) Write(ctx context.Context, wr *prompb.WriteRequest) error {
 }
 
 func (n *nopCloser) CloseIdleConnections() {}
+
+func WriteRequest(ctx context.Context, clients []RemoteWriteClient, wr *prompb.WriteRequest) error {
+	var wg sync.WaitGroup
+	wg.Add(len(clients))
+	errs := make(chan error, len(clients))
+	for _, client := range clients {
+		go func(client RemoteWriteClient) {
```
jessejlt, line 40:
```suggestion
		client := client
		go func(client RemoteWriteClient) {
```
This might not be necessary depending on the Go version, but probably not a bad idea to add these closure captures until we're well beyond the current critical version.

jwilder:
I like it.  Nice find.

jwilder (approved):

jessejlt (approved):
Just one suggestion

PR #547, opened 2025-01-28
(improvement) add maxSegmentAge in healthreporter

vpatelsj:
(improvement) adding a maxSegmentAge field in healthreporter so that it can be tracked in the logs

### DIFF
--- a/metrics/service.go
+++ b/metrics/service.go
@@ -21,6 +21,7 @@ type HealthReporter interface {
 	SegmentsTotal() int64
 	SegmentsSize() int64
 	UnhealthyReason() string
+	MaxSegmentAge() time.Duration
 }
 
 type TimeSeriesWriter interface {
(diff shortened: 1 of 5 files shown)


metrics/service.go
```diff
@@ -142,7 +143,7 @@ func (s *service) collect(ctx context.Context) {
 
 			logger.Infof("Status IngestionSamplesPerSecond=%0.2f SamplesIngested=%d IsHealthy=%v "+
 				"UploadQueueSize=%d TransferQueueSize=%d SegmentsTotal=%d SegmentsSize=%d UnhealthyReason=%s "+
-				"ActiveConnections=%d DroppedConnections=%d",
+				"ActiveConnections=%d DroppedConnections=%d MaxSegmentAgeSeconds=%f",
```
jwilder, line 146: Maybe %0.2f to keep format relatively consistent?

jwilder (changes requested):

metrics/service.go
vpatelsj: thanks, addressed

jwilder (approved):
Look good.

PR #548, opened 2025-01-29
Cleanup ingestor and collector endpoints

mkeesey:
The pattern going forward for performance is to send WAL files to ingestor. Remove the legacy prom remote-write and otlp/grpc logs endpoints from ingestor. Also remove the "proxy" endpoint for oltp/grpc logs from collector, as this is unused and the non proxy impl should be used instead.


jwilder (approved):
Nice.  Removing dead code.

PR #549, opened 2025-01-29
Add proposal for SummaryRule

jwilder:
SummaryRule is a new CRD that allows for summarizing tables or importing data on a schedule.  This is useful for convert some of the raw data with very high cardinality that can be slower to query on larger time intervals into a format that is more efficient to query.

### DIFF
--- a/docs/designs/summary-rules.md
+++ b/docs/designs/summary-rules.md
@@ -1,0 +1,106 @@
+# Summary Rules
+
+## Background
+
+ADX-Mon ingests telemetry into ADX without restrictions on cardinality or dimensionality.  Storing this raw data for long periods of time is expensive and inefficient.  There are times where it would be useful to aggregate this raw data for longer retention, query efficiency or to reduce costs.
+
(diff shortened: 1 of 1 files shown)


docs/designs/summary-rules.md
```diff
@@ -0,0 +1,106 @@
...
+  name: SomeMetricHourlyAvg
+  body: |
+    let _startTime = table('SomeMetricHourlyAvg') | summarize max(Timestamp); 
+    SomeMetric
+    | where Timestamp >= _startTime
+    | summarize avg(Value) by bin(Timestamp, 1h)
+    | union table('SomeMetricHourlyAvg')
+  database: SomeDatabase
+  table: SomeMetricHourlyAvg
+  isView: true
```
jessejlt, line 78: We'll need to make some changes to our Function CRD spec and create a new version to maintain compatibility with existing Functions. I like this approach, though.

jessejlt (approved):

docs/designs/summary-rules.md
jwilder: What changes?  This should be able to use the existing view functionality... or am I missing something?

PR #551, opened 2025-01-29
Protect batch bools with mutex

mkeesey:
These bools are used by tests to determine if batches have been released and/or removed. They are written in a different goroutine than they are read in, so these tests sometimes fail.


ingestor/cluster/batcher.go
```diff
@@ -48,20 +48,28 @@ type Batch struct {
 	Table    string
 	Prefix   string
 
-	batcher  Batcher
+	batcher Batcher
+
 	released bool
 	removed  bool
+	mut      sync.Mutex
```
jwilder, line 55: nit: in other code we name this with just `mu`.

jwilder (approved):

jessejlt (approved):
Looks good

PR #553, opened 2025-01-30
Refactor connection closing logic in limitListenerConn and add tests …

jessejlt:
…for connection limits

If Close is invoked for a LimitListener more than once, the metric tracking the total number of active connections can fall below zero and since this metric is used by Ingestor's health check, Ingestor erroneously thinks it's unhealthy and our readiness probe begins to fail, which in turn is causing our integration tests to fail.


vpatelsj (approved):
lgtm

PR #555, opened 2025-01-30
Create a new input and output queue per journal target

mkeesey:
We create individual batchConfig and workers for each journal source, so we need to create individual channels for this to ensure we don't interfere with one another.

Fixes some panics on shutdown.


jwilder (approved):

jwilder:
Good catch

PR #556, opened 2025-02-01
Generate SummaryRule CRD

jessejlt:
From https://github.com/Azure/adx-mon/blob/main/docs/designs/summary-rules.md

We'll want to discuss where to keep the kubebuilder project so CRDs can be maintained by others. I'm not sure if a subdirectory with the entire kubebuilder scaffold or a different project makes most sense. For now, this PR provides sufficient CRD context to begin building out the feature as described in our design document.


jwilder (approved):
We will probably need to iterate on this a little bit more.  But good to get some initial bit in place.

PR #557, opened 2025-02-03
Create a Functions (CRD) inventory

jessejlt:
Instead of emitting a log for each time we touch a function or a metric with function name and status as dimensions, emit a single log line with function inventory, containing each function's current state. We list functions at a defined cadence and set status and reason fields based on certain criteria, where this PR takes the end result and logs it. This PR also includes a sample log line and answers specific questions using KQL that we might want to ask of the system's state.

Logs are of the form

```json
{"time":"2025-02-03T21:13:37.026341494Z","level":"INFO","msg":"Functions","inventory":[{"name":"new-fn","namespace":"default","reason":"Function created","status":""},{"name":"updated-fn","namespace":"default","reason":"Function updated","status":"Success"}]}
```

Questions and answers

- What functions are currently (last hour) being managed?

```kql
T
| where Timestamp > ago(1h)
| where Body.msg == "Functions"
| mv-expand entry = Body.inventory
| distinct entry.name
```

- What functions are currently (last hour) failing to be created in Kusto?

```kql
T
| where Timestamp > ago(1h)
| where Body.msg == "Functions"
| mv-expand entry = Body.inventory
| where entry.status == "Failed"
| distinct entry.name
```

### DIFF
--- a/ingestor/storage/kql_functions.go
+++ b/ingestor/storage/kql_functions.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 
 	"github.com/Azure/adx-mon/pkg/logger"
 	"github.com/Azure/adx-mon/pkg/scheduler"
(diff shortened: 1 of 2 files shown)


mkeesey:
I don't know that we want to do this - with a bunch of functions the inventory field will get enormous. Wide events are ok, but I think the events themselves should not scale in size to the number of different things triggering events over time. Instead, the wideness can expand based on different dimensions being exposed.

PR #559, opened 2025-02-04
Log forwarded IP for transfer errors

jwilder:
This helps track the source client that may be having issues.


jessejlt (approved):
This will be a big help

PR #560, opened 2025-02-05
(feat) Ingestor shutdown via pod annotation

vpatelsj:
Adding a service shutdown endpoint to the ingestor service. It stops the service from accepting new requests by sending a 503 and waits until all writes have been completed. This can eventually be used by ingestor scaler to scale down service pods safely.


ingestor/service.go
```diff
@@ -343,6 +348,12 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
 		}
 	}()
 
+	if !s.isOpen {
```
jwilder, line 351: I'd use the `DisableWrites()` which will shutdown the HTTP server altogether and close the `Store`.  That will prevent new writes from all paths.  

`isOpen` actually creates a race condition because `HandleTransfer` reads it and `Close` writes it in separate goroutines.  If you add a mutex to protect it, it creates lock contentions and slows write throughput if not done carefully.

jwilder (changes requested):

ingestor/service.go
vpatelsj: DisableWrites() function I only see close on store and metrics service. How does it shutdown http server?

jwilder: My mistake. Looks like the http shutdown is done in `main` just before `DisableWrites` is called.   We could either refactor the Service to include the http server so that `DisableWrites` could call it or use `isOpen` approach.  I'd say use `isOpen` but switch it to use atomics vs member var w/ a mutex.  I'd suggest going with the atomics for now (even though it's still racy), and then we can refactor the server/service separately and switch to that later.

cmd/ingestor/main.go
```diff
@@ -365,6 +365,7 @@ func realMain(ctx *cli.Context) error {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/transfer", svc.HandleTransfer)
 	mux.HandleFunc("/readyz", svc.HandleReady)
+	mux.HandleFunc("/shutdown", svc.HandleShutdown)
```
jwilder, line 368: We can't put this on the exposed API.  You could run a denial of service attack with it.

jwilder (changes requested):

ingestor/service.go
vpatelsj: Reverted to using annotations based invocation on the ingestor pod. Hope this mitigates the denial of attack concerns.

cmd/ingestor/main.go
vpatelsj: No longer exposing an endpoint. Thanks

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+		service:    svc,
+	}
+}
+
+func (r *ShutDownRunner) Run(ctx context.Context) error {
+
+	//get ingestor pod in which this runner is running
+	pod, err := r.k8sClient.CoreV1().Pods(namespace).Get(ctx, os.Getenv("HOSTNAME"), metav1.GetOptions{})
+	if err != nil {
+		logger.Errorf("failed to get pod annotations: %v", err)
```
jwilder, line 39: Just return fmt.Errorf('... %w') since caller logs the error already.

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+	if _, ok := pod.Annotations[SHUTDOWN_COMPLETED]; ok {
+		logger.Infof("shutdown already completed on the pod, skipping shutting down")
+		return nil
+	}
+
+	//shutdown the service
+	if _, ok := pod.Annotations[SHUTDOWN_REQUESTED]; ok {
+		logger.Infof("shutting down the service")
+		if err := r.httpServer.Close(); err != nil {
+			logger.Errorf("failed to close http server: %v", err)
```
jwilder, line 53: Return wrapped error

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+			return err
+		}
+
+		if err := r.service.Shutdown(); err != nil {
+			logger.Errorf("failed to shutdown the service: %v", err)
```
jwilder, line 58: Return wrapped error

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+			return err
+		}
+		logger.Infof("service shutdown completed")
```
jwilder, line 61: Capitalize `Service` for log messages to keep consistent with rest of the code base.

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+		//set shutdown-completed annotation
+		pod.Annotations[SHUTDOWN_COMPLETED] = "true"
+		if _, err := r.k8sClient.CoreV1().Pods(namespace).Update(ctx, pod, metav1.UpdateOptions{}); err != nil {
+			logger.Errorf("failed to set shutdown-completed annotation: %v", err)
```
jwilder, line 65: Return the wrapped error

ingestor/service.go
```diff
@@ -381,6 +391,25 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusAccepted)
 }
 
+func (s *Service) Shutdown() error {
+	if err := s.DisableWrites(); err != nil {
+		logger.Errorf("Failed to disable writes: %s", err.Error())
```
jwilder, line 396: Return wrapped error

ingestor/service.go
```diff
@@ -381,6 +391,25 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
...
+		return err
+
+	}
+
+	if err := s.UploadSegments(); err != nil {
+		logger.Errorf("Failed to upload segments: %s", err.Error())
```
jwilder, line 402: Wrap error

ingestor/service.go
```diff
@@ -381,6 +391,25 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
...
+		return err
+	}
+
+	if err := s.Close(); err != nil {
+		logger.Errorf("Failed to close: %s", err.Error())
```
jwilder, line 407: Wrap error

pkg/scheduler/periodic.go
```diff
@@ -57,3 +57,26 @@ func (s *Periodic) ScheduleEvery(interval time.Duration, name string, fn func(ct
 		}
 	}()
 }
+
+type Runner interface {
+	Run(ctx context.Context) error
+	Name() string
+}
+
+func RunForever(ctx context.Context, interval time.Duration, runners ...Runner) {
```
jwilder, line 66: Can you add docs for how this is different than the existing `ScheduleEvery` function?  Looks like this version will run on every node vs `ScheduleEvery` will only run on the leader.  Both need docs actually.

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+	//get ingestor pod in which this runner is running
+	pod, err := r.k8sClient.CoreV1().Pods(namespace).Get(ctx, os.Getenv("HOSTNAME"), metav1.GetOptions{})
+	if err != nil {
+		logger.Errorf("failed to get pod annotations: %v", err)
+		return err
+	}
+
+	//check if shutdown-completed annotation is set
+	if _, ok := pod.Annotations[SHUTDOWN_COMPLETED]; ok {
+		logger.Infof("shutdown already completed on the pod, skipping shutting down")
```
jwilder, line 45: Capital Shutdown

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+		return nil
+	}
+
+	//shutdown the service
+	if _, ok := pod.Annotations[SHUTDOWN_REQUESTED]; ok {
+		logger.Infof("shutting down the service")
```
jwilder, line 51: Capitalize first word

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+			return err
+		}
+
+		if err := r.service.Shutdown(); err != nil {
+			logger.Errorf("failed to shutdown the service: %v", err)
+			return err
+		}
+		logger.Infof("service shutdown completed")
+		//set shutdown-completed annotation
+		pod.Annotations[SHUTDOWN_COMPLETED] = "true"
```
jwilder, line 63: I think this should only get set after the all uploads have completed?

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,75 @@
...
+	//check if shutdown-completed annotation is set
+	if _, ok := pod.Annotations[SHUTDOWN_COMPLETED]; ok {
+		logger.Infof("shutdown already completed on the pod, skipping shutting down")
+		return nil
+	}
+
+	//shutdown the service
+	if _, ok := pod.Annotations[SHUTDOWN_REQUESTED]; ok {
+		logger.Infof("shutting down the service")
+		if err := r.httpServer.Close(); err != nil {
```
jwilder, line 52: What happens if a shutdown is request and the pods just exits unexpectedly?  How can we make this SHUTDOWN_REQUESTED phase idempotent and resumable?

ingestor/service.go
```diff
@@ -381,6 +391,25 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
...
+		return err
+
+	}
+
+	if err := s.UploadSegments(); err != nil {
+		logger.Errorf("Failed to upload segments: %s", err.Error())
+		return err
+	}
+
+	if err := s.Close(); err != nil {
```
jwilder, line 406: I don't think you want to `Close` here because then the pod will just restart and be able to take new writes.  I think it would be better to have the instance stay in this disabled state and signal back that shudown is complete here.  The operator can then see that pod has signaled shutdown is completed and can delete the pod/scale down the statefulset.

That actually reminds me that when we scale down we have to do it in reverse order because of the way that statefulsets works.

jwilder (changes requested):

ingestor/runner/shutdown/shutdown.go
vpatelsj: The service.Shutdown() has the uploadSegments() and the store.close() call in it already. Is that enough to assume the uploads have been completed?

ingestor/runner/shutdown/shutdown.go
vpatelsj: Annotations are attached to pod object. If the pod exits  unexpectedly but not deleted, the annotations will stay on it. If the same pod starts again, the annotations will still be there and the Runner will pickup the annotations and resume the shutdown.

On the contrary if the pod is recreated, it will not have the annotations. The operator will have to detect if it still wants the new pod object around and then issue another shutdown on it if needed.

ingestor/service.go
vpatelsj: store.close() is called twice once in DisableWrites() and once here. Restructured the code so that we dont call store.close() in shutdown() and not have to call DisableWrites() either because that one only calls metrics.close() besides store.close()

can you elaborate on the reverse order? I am not sure I follow.

jessejlt (approved):

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,72 @@
...
+			return fmt.Errorf("failed to close http server: %v", err)
+		}
+
+		if err := r.service.Shutdown(); err != nil {
+			return fmt.Errorf("failed to shutdown the service: %v", err)
+		}
+		logger.Infof("Service shutdown completed")
+		//set shutdown-completed annotation
+		pod.Annotations[SHUTDOWN_COMPLETED] = "true"
+		if _, err := r.k8sClient.CoreV1().Pods(namespace).Update(ctx, pod, metav1.UpdateOptions{}); err != nil {
```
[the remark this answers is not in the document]
vpatelsj: Good reminder, thanks!

ingestor/runner/shutdown/shutdown.go
```diff
@@ -0,0 +1,72 @@
...
+			fmt.Errorf("failed to set shutdown-completed annotation: %v", err)
```
jwilder, line 63: Missing `return`

ingestor/service.go
```diff
@@ -381,6 +391,18 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusAccepted)
 }
 
+func (s *Service) Shutdown() error {
+	if err := s.metrics.Close(); err != nil {
+		return err
+	}
+
+	if err := s.UploadSegments(); err != nil {
```
jwilder, line 399: I just noticed that this func has a hard coded 30s timeout to upload.  I'd suggest refactoring that to take a `Context` and then you can pass int a `context.WithTimeOut(5 * time.Minutes)` to give it more time to complete uploads.  The 30s timeout is a remnant from the SIGTERM handler that we don't use anymore.

ingestor/runner/shutdown/shutdown.go
jwilder: I don't see the `store.Close` call in there anymore.  I see that `UploadSegments` is synchronous so if it returns, it's finished uploading.  See my other comment about adjust the 30s timeout on there.

jwilder (changes requested):

ingestor/runner/shutdown/shutdown.go
vpatelsj: good catch. thanks

ingestor/service.go
vpatelsj: sounds good. I've added context.Timeout in the uploadWrites method.

ingestor/service.go
```diff
@@ -381,17 +395,29 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func (s *Service) UploadSegments(ctx context.Context) error {
 	if err := s.batcher.BatchSegments(); err != nil {
 		return err
 	}
 	logger.Infof("Waiting for upload queue to drain, %d batches remaining", len(s.uploader.UploadQueue()))
 	logger.Infof("Waiting for transfer queue to drain, %d batches remaining", len(s.replicator.TransferQueue()))
 
+	timeoutCtx, cancel := context.WithTimeout(ctx, shutdownTimeout)
```
jwilder, line 417: Not quite what I was getting at.  I was suggesting to move this part out of `UploadSegments` to let the caller define when to timeout by passing a  context.  So move L417-L418 to the call site.

jwilder (changes requested):

ingestor/service.go
vpatelsj: Gotcha. updated to pass the timeout context from the shutdown runner.

ingestor/runner/shutdown/shutdown.go
vpatelsj: addresssed your timeout request.

jwilder (approved):

PR #561, opened 2025-02-05
Remove some overly verbose debug logging

jwilder:
When DEBUG level is enabled, the logs are flooded with the contents of all the logs collector is shipping which makes them to noisy to be useful.


jessejlt (approved):
Those are my debug logs, sorry about that.

PR #566, opened 2025-02-06
Json parser should fail when not provided json

mkeesey:
Not doing so prevents falling back to other parsers.


jessejlt (approved):
Awesome!

PR #567, opened 2025-02-07
Add custom CodeQL action

tanayk2610:
- This custom action doesn't run integration tests as part of the build step as opposed to the default CodeQL workflow setup using Github. This should reduce our PR build times significantly.
- Updates the `build-collector` Make target as building collector was causing the following error: `go: cannot write multiple packages to non-directory
bin/collector`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ build-ingestor:
 
 build-collector:
 	mkdir -p bin
-	CGO_ENABLED=1 go build -o bin/collector ./cmd/collector/...
+	CGO_ENABLED=1 go build -o bin/collector ./cmd/collector/
 .PHONY: build
 
 build: build-alerter build-ingestor build-collector
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,44 @@
+name: "Custom CodeQL analysis"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  analyze:
+    name: Analyze Go code
+    runs-on: 'ubuntu-latest'
+    permissions:
+      # required for all workflows
+      security-events: write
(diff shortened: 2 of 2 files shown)


jessejlt (approved):
I'm not sure what the make target error is all about. Seems okay to me.

Makefile
```diff
@@ -12,7 +12,7 @@ build-ingestor:
 
 build-collector:
 	mkdir -p bin
-	CGO_ENABLED=1 go build -o bin/collector ./cmd/collector/...
+	CGO_ENABLED=1 go build -o bin/ ./cmd/collector/...
```
jwilder, line 15: This ends up building all 3 binaries which is what the `build` target does.

jwilder: Err... actually, my comment is wrong. Why is this needed?

tanayk2610: This command works:
```bash
CGO_ENABLED=1 go build -o bin/collector ./cmd/collector/
```
This command doesn't:
```bash
CGO_ENABLED=1 go build -o bin/collector ./cmd/collector/...
```
I think it's the config directory which throws `go build` off:
```bash
$ tree cmd/collector
cmd/collector
├── config
│   ├── config.go
│   ├── config_test.go
│   ├── export.go
│   └── export_test.go
├── main.go
└── main_test.go
 ```
 
Here's my hypothesis: go build creates a context around building the necessary binaries under `./cmd/collector/...`, which includes the config directory. Maybe it's just not smart enough to understand that this will just result in one binary and errors out thinking we're trying to create multiple binaries under `bin/collector` which doesn't even exists:
```bash
$ CGO_ENABLED=1 go build -o bin/collector ./cmd/collector/...
go: cannot write multiple packages to non-directory bin/collector
```

jwilder (approved):

Makefile
jwilder: Maybe update the ingestor/alerter targets to remove `...` as well?

PR #572, opened 2025-02-10
OTLP log endpoint converts incoming messages to native log format

mkeesey:
To support future work on unifying log processing throughout collector, this PR converts the OTLP log endpoint to using the "native log" format that our tailers utilize. This will allow us to have a unified format in the future for all log sources for processing, exporting, and converting into the store format to send to ingestor.


jessejlt (approved):
Just one safety suggestion

PR #579, opened 2025-02-13
Reduce the poll interval for function syncing

jessejlt:
Reduce the poll interval to match Alerter. The idea is that we want to discover functions a bit quicker because we don't want Alerter to evaluate a moniter that references a table / view that hasen't yet been created. An alternative approach would be to use an informer, but my experience with informers is that we would need a manual cache hydration anyways, so I think it makes sense to just reduce the poll interval and if the issue persists we can look into adding an informer.


jwilder (approved):
Sounds good.

PR #583, opened 2025-02-14
Add benchmarking tool to interleave benchmark runs

mkeesey:
Enter the directory with your benchmarks, then run bench.sh compile to create the base. With changes, run bench.sh without args to compare the two implementations and print results with benchstat.


jessejlt (approved):
Super helpful!

jwilder (approved):

jwilder (approved):

PR #591, opened 2025-02-24
Enable Conditions subresources for Functions

jessejlt:
We want the ability to perform out-of-band operations against Function state changes, such as applying custom business logic, and track the progress of these operations in a durable way. One way to do this is for the out-of-band process to have its own CRD that references each Function and stores its state therein; however, this is rather heavy-handed just for tracking some purposes. So we instead introduce Conditions to the Function status, which is a common CRD practice for this use case.
This pull request includes several changes to the `api/v1` package and updates to the Kubernetes CustomResourceDefinition (CRD) for functions. The most important changes focus on enhancing the deepcopy functions and adding conditions to the function status.

CRD subresources aren't prescriptive in how they're used, but let's come up with a use case to better understand the intent. Let's say I want to apply some custom business logic in reaction to Functions changing, something like syncing these functions to an alternate location. In this scenario, my non-owner / out-of-band process can react to Function state changes and perform its sync logic, once completed, the non-owner process would update the Function's conditions to reflect its own operations. 

By way of an example, some user creates the following Function and Ingestor creates the Function in the appropriate Kusto cluster

```yaml
apiVersion: adx-mon.azure.com/v1
kind: Function
metadata:
  creationTimestamp: "2025-02-11T22:59:14Z"
  generation: 1
  name: latest-result
  namespace: default
spec:
  body: |
    .create-or-alter function Latest () {
     MyTable
    | summarize max(Timestamp)
    }
  database: MyDatabase
status:
  lastTimeReconciled: "2025-02-11T23:08:50Z"
  observedGeneration: 1
  reason: Function created
  status: Success
```

In response, some other controller that has nothing to do with adx-mon, observes that there is a new Function and applies some business logic, then updates the Function to store reflect that it has already processed the Function

```yaml
apiVersion: adx-mon.azure.com/v1
kind: Function
metadata:
  creationTimestamp: "2025-02-11T22:59:14Z"
  generation: 1
  name: latest-result
  namespace: default
spec:
  body: |
    .create-or-alter function Latest () {
     MyTable
    | summarize max(Timestamp)
    }
  database: MyDatabase
status:
  lastTimeReconciled: "2025-02-11T23:08:50Z"
  observedGeneration: 1
  reason: Function created
  status: Success
  conditions:
  - lastTransitionTime: "2025-02-11T23:08:50Z"
    type: some.business.logic/Handler
    observedGeneration: 1
    message: business logic applied
    reason: FunctionCreated
    status: True
```

Enhancements to deepcopy functions:

* [`api/v1/zz_generated.deepcopy.go`](diffhunk://#diff-4267a35009ae655a91fbca7770788673905c01d1aca230a644b4e76a66d994b8L33-R17): Modified `DeepCopyInto` methods for `AlertRule`, `AlertRuleSpec`, and `Function` to ensure deep copying of nested structures, including handling of `Spec` and `Conditions` fields. [[1]](diffhunk://#diff-4267a35009ae655a91fbca7770788673905c01d1aca230a644b4e76a66d994b8L33-R17) [[2]](diffhunk://#diff-4267a35009ae655a91fbca7770788673905c01d1aca230a644b4e76a66d994b8R76-R91) [[3]](diffhunk://#diff-4267a35009ae655a91fbca7770788673905c01d1aca230a644b4e76a66d994b8L121-R126) [[4]](diffhunk://#diff-4267a35009ae655a91fbca7770788673905c01d1aca230a644b4e76a66d994b8R183-R187) [[5]](diffhunk://#diff-4267a35009ae655a91fbca7770788673905c01d1aca230a644b4e76a66d994b8R204-R210)

Updates to function status:

* [`api/v1/function_types.go`](diffhunk://#diff-081b33660f2560ef3c13ea8b7a03281456a2163f3426d188b35d63c03d5891c2R65-R67): Added `Conditions` field to the `FunctionStatus` struct to track the status conditions of the function.

Kubernetes CRD updates:

* [`kustomize/bases/functions_crd.yaml`](diffhunk://#diff-c595d33142ec65f1413f8edb228268ae86a5fa79872a62089128974bff29020eL6-R6): Updated the CRD version annotation and added a detailed schema for the `Conditions` field in the function status. [[1]](diffhunk://#diff-c595d33142ec65f1413f8edb228268ae86a5fa79872a62089128974bff29020eL6-R6) [[2]](diffhunk://#diff-c595d33142ec65f1413f8edb228268ae86a5fa79872a62089128974bff29020eR62-R119)

Other changes:

* [`api/v1/zz_generated.deepcopy.go`](diffhunk://#diff-4267a35009ae655a91fbca7770788673905c01d1aca230a644b4e76a66d994b8L2-R8): Added import for `metav1` package to support the new `Conditions` field.


jwilder:
This could be useful for the SummaryRules to track last execution state, but since ingestor is the owner, I think it could just update the Status fields too. 

@mkeesey @matucker-msft what do you think about the proposal?

matucker-msft (approved):

PR #594, opened 2025-02-26
Create fluent-forward encoding implementation

mkeesey:
This is not currently hooked up or enabled, but will be used for exporting logs via fluent-forward protocol. This protocol uses msgpack to encode a message that contains a tag and a set of messages. We have implemented forward mode with ext time to carry nanosecond-precision timestamps.

### DIFF
--- a/collector/export/log_fluent.go
+++ b/collector/export/log_fluent.go
@@ -1,0 +1,156 @@
+package export
+
+import (
+	"context"
+	"encoding/binary"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/Azure/adx-mon/collector/logs/types"
+	"github.com/tinylib/msgp/msgp"
+)
+
+// LogToFluentExporter exports logs using the fluent-forward protocol
+type LogToFluentExporter struct {
+	destination string
+}
+
+type LogToFluentExporterOpts struct {
+	Destination string
+}
+
+func (e *LogToFluentExporter) Send(ctx context.Context, batch *types.LogBatch) error {
+	panic("not implemented")
+}
+
+// fluentEncoder encodes logs in the fluentd forward format
+type fluentEncoder struct {
+	fluentExtTime *fluentExtTime
+	w             []byte
+	batchToSend   []*types.Log
+}
+
+func newFluentEncoder() *fluentEncoder {
+	return &fluentEncoder{
+		fluentExtTime: &fluentExtTime{},
+	}
+}
+
+// encode encodes a log batch into the fluentd forward format.
+// Not multi-goroutine safe. Returned bytes are only valid until the next call.
+func (e *fluentEncoder) encode(tagAttribute string, batch *types.LogBatch) ([]byte, error) {
+	// Copy logbatch elements into new slice so we can sort it without modifying the original,
+	// which is shared amongst other outputs.
+	e.batchToSend = e.batchToSend[:0]
+	for _, log := range batch.Logs {
+		_, ok := log.Attributes[tagAttribute].(string)
+		if !ok {
+			continue
+		}
+		e.batchToSend = append(e.batchToSend, log)
+	}
+	if len(e.batchToSend) == 0 {
+		return nil, nil
+	}
+
+	// Sort based on tags. This allows us to create a message per tag.
+	slices.SortFunc(e.batchToSend, func(a, b *types.Log) int {
+		tagOne := a.Attributes[tagAttribute]
+		tagTwo := b.Attributes[tagAttribute]
+		return strings.Compare(tagOne.(string), tagTwo.(string))
+	})
+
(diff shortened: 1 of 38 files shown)


collector/export/log_fluent.go
```diff
@@ -0,0 +1,152 @@
...
+	return 0
+}
+
+func (e *fluentExtTime) Len() int {
+	return 8
+}
+
+// https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1#eventtime-ext-format
+func (e *fluentExtTime) MarshalBinaryTo(b []byte) error {
+	seconds := uint32(e.UnixTsNano / 1_000_000_000)
```
jwilder, line 140: I've never seen this `_` syntax for number before.  I usually use `1e9`.

collector/export/log_fluent.go
```diff
@@ -0,0 +1,152 @@
...
+	nanoSeconds := uint32(e.UnixTsNano % 1_000_000_000)
+	binary.BigEndian.PutUint32(b, uint32(seconds))
+	binary.BigEndian.PutUint32(b[4:], uint32(nanoSeconds))
+	return nil
+}
+
+func (e *fluentExtTime) UnmarshalBinary(b []byte) error {
+	seconds := binary.BigEndian.Uint32(b)
+	nanoSeconds := binary.BigEndian.Uint32(b[4:])
```
jwilder, line 149: This could panic on a nil slice or a slice of len() < 4.

collector/export/log_fluent.go
```diff
@@ -0,0 +1,152 @@
...
+func (e *fluentExtTime) Len() int {
+	return 8
+}
+
+// https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1#eventtime-ext-format
+func (e *fluentExtTime) MarshalBinaryTo(b []byte) error {
+	seconds := uint32(e.UnixTsNano / 1_000_000_000)
+	nanoSeconds := uint32(e.UnixTsNano % 1_000_000_000)
+	binary.BigEndian.PutUint32(b, uint32(seconds))
+	binary.BigEndian.PutUint32(b[4:], uint32(nanoSeconds))
```
jwilder, line 143: Could panic

jwilder (changes requested):

collector/export/log_fluent.go
```diff
@@ -0,0 +1,152 @@
...
+//     partial messages on the wire, which may upset downstream consumers.
+func (e *fluentEncoder) Encode(tagAttribute string, batch *types.LogBatch) ([]byte, error) {
+	// TODO decide if we should reuse this array, count number of matching elements, etc
+
+	// Copy logbatch elements into new slice so we can sort it without modifying the original,
+	// which is shared amongst other outputs.
+	batchToSend := []*types.Log{}
+	for _, log := range batch.Logs {
+		_, ok := log.Attributes[tagAttribute].(string)
+		if !ok {
```
jessejlt, line 51: I realize this is just stubbed out, but for future consideration, we should probably emit some sort of telemetry for this scenario, since it's indicative of a tagging bug somewhere or otherwise undesired and unexpected behavior, as these logs will be dropped.

collector/export/log_fluent.go
mkeesey: They document they always provide a large enough slice (based on `Len()`), but the check is very cheap so I'm adding one.

collector/export/log_fluent.go
```diff
@@ -0,0 +1,152 @@
...
+			err = e.appendMsg(batchToSend[activeTagStart:idx+1], activeTag)
+			if err != nil {
+				return nil, err // skip???
+			}
+			activeTag = peekTag
+			activeTagStart = idx + 1
+		}
+	}
+
+	return e.w, nil
```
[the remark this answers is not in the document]
mkeesey: They will in this case - documented now and I made the method package private to make it more clear about the internal nature of it.

collector/export/log_fluent.go
```diff
@@ -0,0 +1,152 @@
...
+type LogToFluentExporter struct {
+	destination string
+}
+
+type LogToFluentExporterOpts struct {
+	Destination string
+}
+
+func (e *LogToFluentExporter) Send(ctx context.Context, batch *types.LogBatch) error {
+	panic("not implemented")
```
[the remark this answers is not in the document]
mkeesey: For now. Keeping this PR smaller and focused on the encoder right now to make review much easier. This exporter isn't hooked up yet.

collector/export/log_fluent.go
mkeesey: Agreed - I think this will be a good use for traces actually

jwilder (approved):

PR #596, opened 2025-02-28
Add test coverage for alerter service

jwilder:


alerter/service.go
```diff
@@ -170,15 +171,6 @@ func Lint(ctx context.Context, opts *AlerterOpts, path string) error {
 		Tags:        opts.Tags,
 	})
 
-	go func() {
```
jessejlt, line 173: Intentionally removing the metrics endpoint?

jessejlt (approved):

alerter/service.go
jwilder: Yeah.  Lint command doesn't run as a server.  Just a command line task.

PR #598, opened 2025-02-28
Introduce a metric to track the number of bytes we're sending to Kusto

jessejlt:


metrics/metrics.go
```diff
@@ -93,6 +93,13 @@ var (
 		Help:      "Counter of the number of logs uploaded to Kusto",
 	}, []string{"database", "table"})
 
+	IngestorUploadedBytes = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
mkeesey, line 96: I don't think this will quite do what we want. This gauge will get its value overwritten every time it uploads a segment to whatever size a given segment is, so we'll get a lot of different values for this gauge that don't really correspond to any consistent view of how much we're uploading.

I think if we want to keep track of the bytes being uploaded, this needs to be a Counter that we always append to. This way we're keeping track of the total as time goes on, and the multiple goroutines that write the metric won't step on each other and overwrite each other's value.

mkeesey (changes requested):

metrics/metrics.go
jessejlt: Makes sense. Updated to be a counter.

metrics/metrics.go
```diff
@@ -93,6 +93,13 @@ var (
 		Help:      "Counter of the number of logs uploaded to Kusto",
 	}, []string{"database", "table"})
 
+	IngestorUploadedBytes = promauto.NewCounterVec(prometheus.CounterOpts{
```
jwilder, line 96: We also need a metric for bytes received via transfer endpoint.

jessejlt: done

ingestor/service.go
```diff
@@ -370,6 +370,8 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
 		if err := r.Body.Close(); err != nil {
 			logger.Errorf("close http body: %s, path=/transfer duration=%s from=%s", err.Error(), dur.String(), originalIP)
 		}
+
+		metrics.IngestorTransferBytes.Add(float64(r.ContentLength))
```
mkeesey, line 374: I don't believe this will work in most cases. The ContentLength value will often be `-1` because we generally use Chunked Encoding, so the size is unknown. https://pkg.go.dev/net/http#Request.ContentLength

```
Ingestor
| where PreciseTimeStamp > ago(1d)
| where msg contains "slow request: path=/transfer"
| take 2
| project PreciseTimeStamp, msg
```

PreciseTimeStamp	msg
2025-03-11 10:31:45.5195630	slow request: path=/transfer duration=2m31.637263208s from=168.61.240.185:16400 size=-1 file=AKSCCPMetrics_ApiserverRequestDurationSecondsBucket_33qsmriuidnk1_291cd356199b6000.wal
2025-03-11 13:55:24.8308001	slow request: path=/transfer duration=2m31.169581894s from=168.61.240.187:25224 size=-1 file=AKSCCPMetrics_ClusterAutoscalerClusterMemoryCurrentBytes_33qsmriuidnk1_291d305234d5e000.wal

ingestor/adx/uploader.go
```diff
@@ -388,3 +390,32 @@ type clusterDetails struct {
 	ClockDescription       string    `kusto:"ClockDescription"`
 	RuntimeDescription     string    `kusto:"RuntimeDescription"`
 }
+
+// countingReader wraps an io.Reader and counts bytes read without allocations.
+type countingReader struct {
```
jwilder, line 395: Can we move this to something like `pkg/reader`.  Looks general enought to be useful in other places.

ingestor/service.go
jwilder: You can wrap the `r.Body` in a countingReader and record the size counted in a defer.

jwilder (changes requested):

PR #599, opened 2025-02-28
Create read-only views of log/logbatch

mkeesey:
After ingestion and processing of log batches by sources and filters, multiple outputs may receive these batches to send to different endpoints. At this point, it is not safe to modify these Log instances or their properties. To make this more clear and to reduce the chances of this happening, we wrap these objects with read only interfaces.

Go does not support covarience of slices, so it is non-trivial to wrap our existing types without copies to support direct access to the properties inside. Instead, we abstract the iteration and value lookups to allow returning the read only copies as best we can. Go doesn't support the read-only paradigms that some other languages do without expensive copies, so this tries to strike a middle ground between performance and safety.

This has a small performance impact and does require a few allocations because Go cannot ensure that the `any` values within our maps cannot escape our ForEach implementations. For the encoding benchmarks for fluent-foward protocol, this generates ~96 bytes per call over 6 allocations, significantly smaller than many of our other implementations in Collector.


jessejlt:
Can't we just introduce a read-only bool to the logs struct and consult the bool before attempting mutation?

PR #601, opened 2025-03-03
Support conveying journal fields through logs

jessejlt:
We need the ability to optionally record [journal fields](http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html) in our logs. I've chosen to record journal fields in the log's `Resource`, because it's meant to communicate entity-level metadata, such as a service's name, so in this case we might want to communicate a systemd's unit or identifier name.


jwilder (approved):
We need to get these added to the docs pages as well.

PR #605, opened 2025-03-05
Introduce SummaryRules

jessejlt:
From our proposal documentation https://github.com/Azure/adx-mon/blob/main/docs/designs/summary-rules.md

> ADX-Mon ingests telemetry into ADX without restrictions on cardinality or dimensionality. Storing this raw data for long periods of time is expensive and inefficient. There are times where it would be useful to aggregate this raw data for longer retention, query efficiency or to reduce costs.
> 
> Similarly, sometimes there is data in other ADX clusters that would be useful to join with the local telemetry collected by ADX-Mon. While this data can be queried using cluster() functions, sometimes the cluster is geographically far from the local cluster and this approach is not as performant as having the data locally. In addition, the remote cluster may not have the same retention policies as the local cluster which can lead to query issues.

This PR factors our `ingestor/storage` into a common CRDHandler for ManagementCommands and SumaryRules. Our Functions can also be refactored to use CRDHandler but doing so is a little more tricky because of the way `Status.Conditions` are handled in Functions. In a future PR I'll factor Functions to use the common handler. 

In addition, I'm becoming concerned with the amount of apiserver listings that are occurring as a result of the way we execute our CRD handlers. In a subsequent PR I'm going to make use of `controller-manager` caching mechanisms to reduce the number of listings that are occurring.

This pull request introduces several changes to the `api/v1`, `ingestor`, and `storage` packages to enhance the handling and processing of custom resource definitions (CRDs) and their conditions. The most important changes include the addition of new interfaces and methods for managing conditions, updates to the `ManagementCommand` and `SummaryRule` types, and the implementation of new tasks for handling summary rules.

### Enhancements to CRD Handling:

* [`api/v1/conditions.go`](diffhunk://#diff-cd57264229aa9d98795ef47e4968bc07cd298def96ade4df6784cd15a7fbb8b5R1-R10): Introduced the `ConditionedObject` interface with `GetCondition` and `SetCondition` methods for managing conditions on CRDs.
* [`ingestor/storage/crds.go`](diffhunk://#diff-0459fd51666979fe00992ea917f060d24cebae11f653678aaa64914b5699f106R1-R165): Added a new `CRDHandler` interface and its implementation for listing and updating the status of CRDs with condition handling.

### Updates to ManagementCommand and SummaryRule Types:

* [`api/v1/managementcommand_types.go`](diffhunk://#diff-d3daa5f89872ba589b2e9fb599e40a0d28c9422915eab8c58136c3d719cd5523R4-R10): Added `ManagementCommandConditionOwner` constant and implemented `GetCondition` and `SetCondition` methods for `ManagementCommand`. [[1]](diffhunk://#diff-d3daa5f89872ba589b2e9fb599e40a0d28c9422915eab8c58136c3d719cd5523R4-R10) [[2]](diffhunk://#diff-d3daa5f89872ba589b2e9fb599e40a0d28c9422915eab8c58136c3d719cd5523R36-R54)
* [`api/v1/summaryrule_types.go`](diffhunk://#diff-b9f68ec13d567ad34a5c6755243c97e090fa0f36442b0fe67d4c81b3aacd9601R20-R26): Added `SummaryRuleOwner` constant, updated the `SummaryRuleStatus` struct, and implemented `GetCondition` and `SetCondition` methods for `SummaryRule`. [[1]](diffhunk://#diff-b9f68ec13d567ad34a5c6755243c97e090fa0f36442b0fe67d4c81b3aacd9601R20-R26) [[2]](diffhunk://#diff-b9f68ec13d567ad34a5c6755243c97e090fa0f36442b0fe67d4c81b3aacd9601L49-R77)

### Implementation of New Tasks:

* [`ingestor/adx/tasks.go`](diffhunk://#diff-86682be9a1330c75a309ebc3ae6c0e02c0baa53bfeb9b9ff98aea176ea3dea4fR8-R18): Introduced `SummaryRuleTask` for executing summary rules and updating their conditions. Updated `ManagementCommandTask` to use the new `CRDHandler` interface. [[1]](diffhunk://#diff-86682be9a1330c75a309ebc3ae6c0e02c0baa53bfeb9b9ff98aea176ea3dea4fR8-R18) [[2]](diffhunk://#diff-86682be9a1330c75a309ebc3ae6c0e02c0baa53bfeb9b9ff98aea176ea3dea4fL193-R212) [[3]](diffhunk://#diff-86682be9a1330c75a309ebc3ae6c0e02c0baa53bfeb9b9ff98aea176ea3dea4fL218-R301)

### Test Enhancements:

* [`ingestor/adx/tasks_test.go`](diffhunk://#diff-e984fe8711d099d1f3fe28c5fd4b3ea01a814a4ae3d41ebe27783ad00ba25147R320-R415): Added integration tests for `SummaryRuleTask` to verify the execution and condition updates of summary rules.

### Miscellaneous Changes:

* [`ingestor/service.go`](diffhunk://#diff-cdc4c0859afc8f4c1fa3efb19ee6d9f8d57fbcf6a553951683735670d7d016d4R267): Updated the service to use the new `CRDHandler` for scheduling management command tasks. [[1]](diffhunk://#diff-cdc4c0859afc8f4c1fa3efb19ee6d9f8d57fbcf6a553951683735670d7d016d4R267) [[2]](diffhunk://#diff-cdc4c0859afc8f4c1fa3efb19ee6d9f8d57fbcf6a553951683735670d7d016d4R278-R294)


ingestor/adx/tasks.go
```diff
@@ -215,16 +218,84 @@ func (t *ManagementCommandTask) Run(ctx context.Context) error {
...
+	if err := t.store.List(ctx, summaryRules); err != nil {
+		return fmt.Errorf("failed to list summary rules: %w", err)
+	}
+	for _, rule := range summaryRules.Items {
+		if rule.Spec.Database != t.kustoCli.Database() {
+			continue
+		}
+		if rule.DeletionTimestamp != nil {
+			// rule is being deleted, we might possibly want to delete the associated
+			// destination table, but there is a potential for unintended historical
```
jwilder, line 257: Definitely don't think we should delete the table by deleting the rule.  Deleting the table should occur though separate affirmative action.

ingestor/adx/tasks.go
```diff
@@ -215,16 +218,84 @@ func (t *ManagementCommandTask) Run(ctx context.Context) error {
...
+			}
+			continue
+		}
+
+		if cnd == nil {
+			cnd = &metav1.Condition{}
+		}
+
+		// Execution occurs asynchronously, so we don't have to concern ourselves with blocking
+		startTime := time.Now().Add(-rule.Spec.Interval.Duration)
```
jwilder, line 275: We need to make these start and end time fall on regular intervals.  I think we should have a minimum of minutely boundaries so we don't allow running something every few seconds faster.  We'd need the start and end times to zero out all the second, millis, nanos.

ingestor/adx/tasks.go
```diff
@@ -215,16 +218,84 @@ func (t *ManagementCommandTask) Run(ctx context.Context) error {
...
+		endTime := metav1.Now()
+		if !cnd.LastTransitionTime.IsZero() {
+			startTime = cnd.LastTransitionTime.Time
+		}
+
+		// Save our endTime, which will become the next iteration's startTime
+		cnd.LastTransitionTime = endTime
+		rule.SetCondition(*cnd)
+
+		// NOTE: We cannot do something like `let _startTime = datetime();` as dot-command do not permit
```
jwilder, line 285: Really?  huh.  That's unfortunate.

ingestor/adx/tasks.go
```diff
@@ -215,16 +218,84 @@ func (t *ManagementCommandTask) Run(ctx context.Context) error {
...
+		// preceding let-statements.
+		rule.Spec.Body = strings.ReplaceAll(rule.Spec.Body, "_startTime", fmt.Sprintf("datetime(%s)", startTime.UTC().Format(time.RFC3339Nano)))
+		rule.Spec.Body = strings.ReplaceAll(rule.Spec.Body, "_endTime", fmt.Sprintf("datetime(%s)", endTime.UTC().Format(time.RFC3339Nano)))
+		// Execute asynchronously
+		stmt := kql.New(".set-or-append async ").AddUnsafe(rule.Spec.Table).AddLiteral(" <| ").AddUnsafe(rule.Spec.Body)
```
jwilder, line 290: Not necessary for this PR, but we may need to track the async request and only update the status if it's successful.  A workaroudn would be to execute this synchronously for now and see if it completes.  That might require some tweaks to the scheduler to better support longer running jobs though.

ingestor/adx/tasks.go
```diff
@@ -215,16 +218,84 @@ func (t *ManagementCommandTask) Run(ctx context.Context) error {
...
+		_, err := t.kustoCli.Mgmt(ctx, stmt)
+		if err != nil {
+			logger.Errorf("Failed to execute summary rule %s.%s: %v", rule.Spec.Database, rule.Name, err)
```
jwilder, line 293: Should we update the status with error details?

ingestor/storage/crds.go
```diff
@@ -0,0 +1,165 @@
...
+		return errors.New("no client provided")
+	}
+
+	if err := c.Client.List(ctx, list); err != nil {
+		logger.Error("Failed to list CRDs: %v", err)
+		return err
+	}
+
+	var filtered []runtime.Object
+	meta.EachListItem(list, func(item runtime.Object) error {
```
jwilder, line 73: Use `_ = ...` to explicitly acknowledge that we don't care about the error?

ingestor/storage/crds.go
```diff
@@ -0,0 +1,165 @@
...
+
+func NewCRDHandler(client client.Client, elector scheduler.Elector) CRDHandler {
+	return &crdHandler{
+		Client:  client,
+		Elector: elector,
+	}
+}
+
+func (c *crdHandler) List(ctx context.Context, list client.ObjectList, filters ...ListFilterFunc) error {
+	if c.Elector != nil && !c.Elector.IsLeader() {
```
jwilder, line 51: I think we'll want to change this (not now) to have the tasks run on all nodes all the time, but partition the tasks to determine which node is the owner and responsible for running the tasks.  Not sure if any of these tasks will incur resource usage but it will let us spread the load more evenly.

ingestor/storage/crds.go
```diff
@@ -0,0 +1,165 @@
...
+	// mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{})
+	// then whenever you retrieve a client, you intherit the caching mechanism
+	// client := mgr.GetClient()
+
+	if c.Client == nil {
+		return errors.New("no client provided")
+	}
+
+	if err := c.Client.List(ctx, list); err != nil {
+		logger.Error("Failed to list CRDs: %v", err)
```
jwilder, line 68: Just return wraped error and let a higher level caller log it.

ingestor/storage/crds.go
```diff
@@ -0,0 +1,165 @@
...
+	condition := metav1.Condition{
+		Status:  status,
+		Message: message,
+	}
+
+	statusObj.SetCondition(condition)
+	logger.Infof("Updating status for %s/%s: %s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName(), message)
+
+	if err := c.Client.Status().Update(ctx, obj); err != nil {
+		logger.Errorf("Failed to update status for %s/%s: %v", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName(), err)
```
jwilder, line 119: Return a wrapped error instead of logging at this level.

jwilder (changes requested):

ingestor/adx/tasks.go
jessejlt: It is on line 296

> t.store.UpdateStatus(ctx, &rule, err)

Where the error is serialized into the status condition

ingestor/adx/tasks.go
jessejlt: The way we report status is through CRD subresources, which means there is a list of Condition objects, each of which has an owner field. Right now, we own only a single Condition, which is tracking the execution times and run status for each execution. Perhaps we could introduce a second Condition, owned by us, that tracks the longer running async operation and reports its status. So something like

```
for each task execution
  if task successfully ran async
    async_status = get async operation result
    if async_status is complete
      set task.Status.Condition[async-operation].Status =  async_status.Succeeded ? v1.ConditionTrue : v1.ConditionFalse
```

ingestor/storage/crds.go
jessejlt: done

ingestor/adx/tasks.go
jwilder: I was highlighting a different issue.  I was saying that the `async` aspect won't let us know if the query actually succeeded or had an ingestion failure or something.   We can track it like you suggested, but just saying that right now we have no code checking on the async status.

jwilder (approved):

PR #609, opened 2025-03-06
Disk usage tuning

jwilder:
This pull request introduces a new feature to limit the number of segments in a batch and includes several changes to improve the handling and testing of segments. The most important changes are summarized below:

### New Feature: Max Batch Segments

* Added a new flag `max-batch-segments` to the CLI in `cmd/ingestor/main.go` to specify the maximum number of segments per batch. [[1]](diffhunk://#diff-5ab380573a6eb9c62fc65adafbeb865ad5677e5aab550a7dfe9bf65b4e0ab711R65) [[2]](diffhunk://#diff-5ab380573a6eb9c62fc65adafbeb865ad5677e5aab550a7dfe9bf65b4e0ab711R131) [[3]](diffhunk://#diff-5ab380573a6eb9c62fc65adafbeb865ad5677e5aab550a7dfe9bf65b4e0ab711R350)
* Updated `ServiceOpts` in `collector/service.go` and `ingestor/service.go` to include `MaxBatchSegments` and propagate this option through the service initialization. [[1]](diffhunk://#diff-dcf3a410ebc2ab76b38109ea1a97c67b5e22bf0ae3c0a93491bc2bd4509f5708R128-R133) [[2]](diffhunk://#diff-dcf3a410ebc2ab76b38109ea1a97c67b5e22bf0ae3c0a93491bc2bd4509f5708R320) [[3]](diffhunk://#diff-cdc4c0859afc8f4c1fa3efb19ee6d9f8d57fbcf6a553951683735670d7d016d4R142-R147) [[4]](diffhunk://#diff-cdc4c0859afc8f4c1fa3efb19ee6d9f8d57fbcf6a553951683735670d7d016d4R193)
* Modified `BatcherOpts` and `batcher` in `ingestor/cluster/batcher.go` to handle the new `MaxBatchSegments` option and ensure batches do not exceed this limit. [[1]](diffhunk://#diff-ae24453fcee063ab87b503a7b3642158e146498522548702a829cb2a3534e921R34) [[2]](diffhunk://#diff-ae24453fcee063ab87b503a7b3642158e146498522548702a829cb2a3534e921R143) [[3]](diffhunk://#diff-ae24453fcee063ab87b503a7b3642158e146498522548702a829cb2a3534e921R155-R166) [[4]](diffhunk://#diff-ae24453fcee063ab87b503a7b3642158e146498522548702a829cb2a3534e921L345-R372)

### Segment Handling Improvements

* Changed the `Append` method in `pkg/wal/segment.go` to return the number of bytes written and updated related tests. [[1]](diffhunk://#diff-0a73f280a85e0976124bdf316c9454fd240d2d33f2899335fdbfea0d49373eb2L55-R60) [[2]](diffhunk://#diff-0a73f280a85e0976124bdf316c9454fd240d2d33f2899335fdbfea0d49373eb2L329-R355) [[3]](diffhunk://#diff-779ec5efee625ad7bf27ca8f8d89289998b432339282b4e849732325894ac804L291-R293) [[4]](diffhunk://#diff-779ec5efee625ad7bf27ca8f8d89289998b432339282b4e849732325894ac804L325-R327) [[5]](diffhunk://#diff-779ec5efee625ad7bf27ca8f8d89289998b432339282b4e849732325894ac804L368-R375) [[6]](diffhunk://#diff-779ec5efee625ad7bf27ca8f8d89289998b432339282b4e849732325894ac804L416-R424)
* Simplified the `Size` method in `pkg/wal/segment.go` to directly return the size without error handling. [[1]](diffhunk://#diff-0a73f280a85e0976124bdf316c9454fd240d2d33f2899335fdbfea0d49373eb2L278-R279) [[2]](diffhunk://#diff-0a73f280a85e0976124bdf316c9454fd240d2d33f2899335fdbfea0d49373eb2L295-R295)

### WAL Enhancements

* Introduced new fields in the `WAL` struct in `pkg/wal/wal.go` to track inflight write bytes, segment size, and segment creation time to reduce lock contention.
* Added a new error `ErrMaxSegmentSizeExceeded` and updated the `validateLimits` method to check for segment size limits. [[1]](diffhunk://#diff-d1376e7d1b7d397f29bcda9a052cd90550cd8f4139fe3f89657070b1c56b3c7dR24) [[2]](diffhunk://#diff-d1376e7d1b7d397f29bcda9a052cd90550cd8f4139fe3f89657070b1c56b3c7dL171-R185) [[3]](diffhunk://#diff-d1376e7d1b7d397f29bcda9a052cd90550cd8f4139fe3f89657070b1c56b3c7dL203-R229)
* Refactored the segment rotation logic into a separate method `rotateSegmentIfNecessary` to improve readability.

### Tests

* Added a new test `TestBatcher_MaxSegmentCount` in `ingestor/cluster/batcher_test.go` to verify the behavior when the maximum number of segments per batch is reached.

These changes collectively enhance the system's ability to manage segment batching and improve the performance and reliability of segment handling.


ingestor/cluster/batcher.go
```diff
@@ -150,11 +152,18 @@ func NewBatcher(opts BatcherOpts) Batcher {
 	if minUploadSize == 0 {
 		minUploadSize = 100 * 1024 * 1024 // This is the minimal "optimal" size for kusto uploads.
 	}
+
+	maxBatchSegments := 25
```
jessejlt, line 156: Just a suggestion, but I could see how years down the road looking at this value and asking "why, why 25?". I know by reviewing this that 25 is the default value, but perhaps to codify that this is a default, we should create something like

```go
const DefaultMaxBatchSegments = 25
```

ingestor/cluster/batcher.go
```diff
@@ -342,7 +351,25 @@ func (b *batcher) processSegments() ([]*Batch, []*Batch, error) {
 				return n + 1, nil
 			})
 
-			// The batch is at the optimal size for uploading to kusto, upload directly and start b new batch.
+			// Prevent trying to combine an unbounded number of segments at once even if they are very small.  This
+			// can incur a log of CPU time and slower transfer when there are hundreds of segments that can be combined.
```
jessejlt, line 355:
```suggestion
			// can incur a lot of CPU time and slower transfers when there are hundreds of segments that can be combined.
```

ingestor/service.go
```diff
@@ -139,6 +139,12 @@ type ServiceOpts struct {
 
 	// EnableWALFsync enables fsync of segments before closing the segment.
 	EnableWALFsync bool
+
+	// MaxBatchSegments is the maximum number of segments to include when transferring segments in a batch.  The segments
+	// are merged into a new segment.  A higher number takes longer to combine on the sending size and increases the
```
jessejlt, line 144:
```suggestion
	// are merged into a new segment.  A higher number takes longer to combine on the sending side and increases the
```

ingestor/service.go
```diff
@@ -139,6 +139,12 @@ type ServiceOpts struct {
...
+	// size of segments on the receiving size.  A lower number creates more batches and high remote transfer calls.  If
```
jessejlt, line 145:
```suggestion
	// size of segments on the receiving side.  A lower number creates more batches and high remote transfer calls.  If
```

pkg/wal/wal.go
```diff
@@ -237,58 +267,54 @@ func (w *WAL) rotate(ctx context.Context) {
...
-					WithFsync(w.opts.EnableWALFsync))
-				if err != nil {
-					logger.Errorf("Failed to create new segment: %s", err.Error())
-					w.segment = nil
-				}
-			}
+		w.mu.Lock()
+		// Re-verify rotation is needed under write lock since the fast path check is racy
+		if (w.opts.SegmentMaxSize > 0 && atomic.LoadInt64(&w.segmentSize) < w.opts.SegmentMaxSize) &&
+			(w.opts.SegmentMaxAge.Seconds() > 0 && time.Since(time.Unix(w.segmentCreatedAt, 0)) < w.opts.SegmentMaxAge) {
```
jessejlt, line 282: Since the fast-path check is the same as this "real check", does it make sense to put this in some sort of predicate to ensure these two checks are kept in sync? Thinking about the engineer that tries to change this logic a year from now and doesn't realize there is a duplicate block that requires updating. Perhaps a comment as to the duplicate nature at a minimum.

pkg/wal/wal.go
```diff
@@ -316,17 +342,28 @@ func (w *WAL) Remove(path string) error {
...
-		return w.tryAppend(ctx, buf)
+	atomic.AddInt64(&w.inflightWriteBytes, int64(len(buf)))
+	defer atomic.AddInt64(&w.inflightWriteBytes, -int64(len(buf)))
+
+	n, err := w.tryAppend(ctx, buf)
+	if errors.Is(err, ErrMaxSegmentSizeExceeded) {
+		w.rotateSegmentIfNecessary()
+		n, err = w.tryAppend(ctx, buf)
+		atomic.AddInt64(&w.segmentSize, int64(n))
+		return err
```
jessejlt, line 353: if `tryAppend` returns no error, it will overwrite the returned error. 

```suggestion
		n, appendErr := w.tryAppend(ctx, buf)
		atomic.AddInt64(&w.segmentSize, int64(n))
		return errors.Join(err, appendErr)
```

pkg/wal/wal.go
```diff
@@ -316,17 +342,28 @@ func (w *WAL) Remove(path string) error {
...
+	} else if errors.Is(err, ErrSegmentClosed) {
+		n, err = w.tryAppend(ctx, buf)
+		atomic.AddInt64(&w.segmentSize, int64(n))
+		return err
```
jessejlt, line 357: same as above, shadowed error

pkg/wal/wal.go
jwilder: This intended because we are handling the ErrMaxSegmentSizeExceeded error here.  When that error is detected, we rotate the segment.  We don't want to return the ErrMaxSegmentSizeExceeded error.

pkg/wal/wal.go
jwilder: This is intended because we are handling the `ErrSegmentClosed` case.

pkg/wal/wal.go
jwilder: These are not quite the same as written.  These two checks are the inverse of each other.  I can rework this though.

ingestor/cluster/batcher.go
jwilder: Fixed

ingestor/service.go
jwilder: Fixed

ingestor/cluster/batcher.go
jwilder: Done

ingestor/service.go
jwilder: Fixed

pkg/wal/wal.go
jwilder: Updated

jessejlt (approved):

PR #615, opened 2025-03-07
Track SummaryRule's async operations

jessejlt:
Tracks SummaryRule's async operation status in Kusto and reflect the status in the SummaryRule's subresources as a new Status Condition. Also handles resubmitting the async operation if the Kusto status states that the job should be retried.

This pull request introduces several enhancements and new functionalities to the `SummaryRule` handling in the `api/v1/summaryrule_types.go` and `ingestor/adx/tasks.go` files. The primary focus is on adding support for asynchronous operations and improving the overall robustness of the `SummaryRule` execution.

### Enhancements to `SummaryRule` handling:

* **Asynchronous Operations Support:**
  - Added constants for `SummaryRuleOperationIdOwner` and `SummaryRuleAsyncOperationPollInterval` to manage asynchronous operations.
  - Introduced `GetOperationIDCondition` and `SetOperationIDCondition` methods to manage the operation ID condition.
  - Implemented `pollAsyncOperation` method to periodically check the status of asynchronous operations and handle their results.

* **Condition Handling Improvements:**
  - Enhanced the `Run` method to handle the new asynchronous operation logic, including setting and updating the operation ID condition.
  - Added `ShouldSummaryRule` and `ShouldSummaryRuleAsync` helper functions to determine when to execute or poll a `SummaryRule`.

### Testing Enhancements:

* **Test Adjustments:**
  - Updated tests in `tasks_test.go` to include checks for the new asynchronous operation conditions and ensure proper handling.
  - Added necessary imports for the new meta package in `tasks_test.go`.


ingestor/adx/tasks.go
```diff
@@ -288,14 +294,274 @@ func (t *SummaryRuleTask) Run(ctx context.Context) error {
...
 }
+
+func (t *SummaryRuleTask) pollAsyncOperation(ctx context.Context, rule v1.SummaryRule) error {
+	cnd := rule.GetOperationIDCondition()
+	if cnd == nil {
+		// We shouldn't ever hit this unless the ShouldSummaryRuleAsync predicate is incorrect.
+		return fmt.Errorf("no operation ID condition found for summary rule %s.%s", rule.Spec.Database, rule.Name)
+	}
+
+	stmt := kql.New(".show operations").
```
jwilder, line 334: The number of queries this will incur will scale 1:1 with the number of SummaryRules that have been executed.  We can do this more efficiently with one query on an interval that returns all operations that have complated and then update each status for them.  That will make this code always run 1 query regardless of how many SummaryRule operation IDs need to be checked.

ingestor/adx/tasks.go
```diff
@@ -252,18 +252,24 @@ func (t *SummaryRuleTask) Run(ctx context.Context) error {
...
+			// If we do not need to execute the SummaryRule as determined
+			// by the predicate, we know the SummaryRule has successfully
+			// been submitted to Kusto and is up-to-date as determined
+			// by the CRD's execution interval.
 			if logger.IsDebug() {
 				logger.Debugf("Skipping summary rule %s.%s as it was executed recently %s", rule.Spec.Database, rule.Name, cnd.LastTransitionTime)
 			}
+
+			// At this point the SummaryRule has successfully been submitted to
+			// Kuto, now we check if it's necessary to poll the async operation.
```
jwilder, line 267: s/Kuto/Kusto/

ingestor/adx/tasks.go
```diff
@@ -288,14 +294,274 @@ func (t *SummaryRuleTask) Run(ctx context.Context) error {
...
+		// https://learn.microsoft.com/en-us/kusto/management/show-operations
+		switch status.State {
+		case "Completed":
+			// Operation succeeded
+			cnd.Status = metav1.ConditionTrue
+			cnd.Reason = "Succeeded"
+		case "Failed":
+			// If the operation failed, we need to check if we should retry
+			if status.ShouldRetry {
+				// Kuso thinks it's worth retrying. We'll set the management submission
```
jwilder, line 380: Kusto

ingestor/adx/tasks.go
```diff
@@ -288,14 +294,274 @@ func (t *SummaryRuleTask) Run(ctx context.Context) error {
...
+				// condition to False, which will have the effect of re-submitting the command.
```
jwilder, line 381: Will it resubmit with the exact same start/end time?  These need to be precisely executed to prevent issues with data rollups at the time boundaries.

ingestor/adx/tasks.go
```diff
@@ -288,14 +294,274 @@ func (t *SummaryRuleTask) Run(ctx context.Context) error {
...
+				cnd.Reason = "Failed"
+			}
+
+		case "Scheduled", "InProgress":
+			// Just record what's happening and we'll check again later.
+			cnd.Reason = status.State
+
+		case "Throttled":
+			// NOTE:
+			// We're being throttled, so this async operation will not be retried
```
jwilder, line 406: This seems like we should retry later.

ingestor/adx/tasks.go
```diff
@@ -288,14 +294,274 @@ func (t *SummaryRuleTask) Run(ctx context.Context) error {
...
+	// For reasons I don't understand, the GO SDK is not able to execute the query
+	// `.show operations | where OperationId == "operationId" | summarize arg_max(LastUpdatedOn, State, ShouldRetry)`.
+	// The client library itself throws an error stating that ShouldRetry should be a bool but is a float64.
+	// Even if I cast ShouldRetry to a bool, the error persists.
+	//
+	// This method is a work-around where we query if the operation should be retried.
+	// `.show operations | where OperationId == "operationId" | where State == "state" | where ShouldRetry == true | count`
+	// If the value is non zero, we should retry.
+	stmt := kql.New(".show operations").
+		AddLiteral(" | where OperationId == ").AddString(operationId).
```
jwilder, line 444: This should be bulk oriented vs per single operation so it doesn't spawn tons of queries if many need to be retried.

jwilder (changes requested):

ingestor/adx/tasks.go
jessejlt: This is a good point. I'll have to invert the relationship. Right now the controllers receive some list of SummaryRules and updates each one based on the operations. I'll update this to instead poll the operations then fetch the SummaryRules and update accordingly such that the operations drive the SummaryRules instead of the other way around.

jessejlt: This is now updated. The flow works as follows:

```
for each task execution interval
  fetch all async operations from Kusto limited by 1d or 5k records
    for each SummaryRule
      if the SummaryRule needs to be executed due to various reasons like its specified interval having elapsed
        execute the SummaryRule and store its async operation id
      for each of the async operations stored in the SummaryRule
        if the async operation has been completed by Kusto, remove it from the outstanding async operations in the Rule
        if the async operation has failed or is retriable, resubmit with the original start and end times
```

So we store historical records of all async operations for each SummaryRule, removing the historical instances once they are completed in Kusto. Any that need to be retried are resubmitted with the originally submitted start and end times, which are stored on the historic record object.

ingestor/adx/tasks.go
```diff
@@ -243,59 +243,211 @@ func NewSummaryRuleTask(store storage.CRDHandler, kustoCli StatementExecutor) *S
...
+		return "", fmt.Errorf("failed to execute summary rule %s.%s: %w", rule.Spec.Database, rule.Name, err)
+	}
+
+	return operationIDFromResult(res)
+}
+
+func (t *SummaryRuleTask) GetOperations(ctx context.Context) ([]AsyncOperationStatus, error) {
+	// List all the async operations that have been executed in the last 24 hours. If one of our
+	// async operations falls out of this window, it's time to stop trying that particular operation.
+	stmt := kql.New(".show operations | where StartedOn > ago(1d) | where Operation == 'TableSetOrAppend' | summarize arg_max(LastUpdatedOn, OperationId, State, ShouldRetry) by OperationId | project LastUpdatedOn, OperationId = tostring(OperationId), State, ShouldRetry = todouble(ShouldRetry)")
```
jwilder, line 389: If there are multiple pending operations for the same summary rule,  will they be applied in order?

jwilder (approved):

ingestor/adx/tasks.go
jessejlt: Yes

ingestor/adx/tasks.go
jessejlt: Added a sort

PR #619, opened 2025-03-11
Enable distributed CRD processing

jessejlt:
There are two primary changes in this PR. 

1. Use the intended controller-manager's caching mechanism to prevent redundant list invocations. 
2. Support leaderless distributed processing of our CRDs, of which there can be a great many. 

The first change isn't really worth explanation, it's an optimization and the intended use of a controller-manager client.

The second is more interesting. We're leveraging an established pattern that enables leaderless distributed processing of CRDs while ensuring only one controller processes any given CRD at a time. 

We introduce two annotations, one to track ownership and one to track the last updated time. 
1. If a CRD is not owned, controller(s) will attempt to claim ownership by setting the necessary annotations. Kubernetes guarantees that only one controller can set these annotations due to the immutable ResourceVersion on the CRDs.
2. If a CRD is not owned by a controller but the last updated time is > 5 minutes, we consider the CRD to be stale, either due to a controller no longer existing or otherwise not making forward progress. In this case we fall back to step 1, where we attempt to claim owership of the CRD.
3. If a CRD is not owned by a controller and has been updated recently, the CRD is ignored by the controller during the current iteration. 
4. If a CRD is owned by a controller it is processed.

This pull request introduces several changes to the `ingestor` package to enhance the handling of controller ownership and improve the testing framework. The key changes involve the addition of ownership annotations, the removal of leader election logic, and the introduction of new tests for ownership handling.

### Enhancements to Ownership Handling:
* Added `ControllerOwnerKey` and `LastUpdatedKey` annotations to track ownership and update times for resources in `api/v1/groupversion_info.go`.
* Implemented a `CheckOwnership` function in `ingestor/storage/crds.go` to verify and claim resource ownership, ensuring that controllers can manage resources effectively.
* Modified `NewCRDHandler` and `NewFunctions` constructors to initialize `ControllerId` instead of using an elector. [[1]](diffhunk://#diff-0459fd51666979fe00992ea917f060d24cebae11f653678aaa64914b5699f106L40-L62) [[2]](diffhunk://#diff-53ca3f64d88bbe9e987f96c6b1f85fd8820b53b97255b51d1471f425d0c5e873L30-R36)

### Codebase Simplification:
* Removed leader election logic from `ingestor/storage/crds.go` and `ingestor/storage/kql_functions.go` to simplify the codebase. [[1]](diffhunk://#diff-0459fd51666979fe00992ea917f060d24cebae11f653678aaa64914b5699f106R66-R84) [[2]](diffhunk://#diff-53ca3f64d88bbe9e987f96c6b1f85fd8820b53b97255b51d1471f425d0c5e873L70-L73)
* Updated various test files to reflect the removal of leader election and to use the new constructors. [[1]](diffhunk://#diff-e984fe8711d099d1f3fe28c5fd4b3ea01a814a4ae3d41ebe27783ad00ba25147L180-R180) [[2]](diffhunk://#diff-e984fe8711d099d1f3fe28c5fd4b3ea01a814a4ae3d41ebe27783ad00ba25147L321-R321) [[3]](diffhunk://#diff-e984fe8711d099d1f3fe28c5fd4b3ea01a814a4ae3d41ebe27783ad00ba25147L419-R419) [[4]](diffhunk://#diff-cdc4c0859afc8f4c1fa3efb19ee6d9f8d57fbcf6a553951683735670d7d016d4L270-R271) [[5]](diffhunk://#diff-6b42ceee599d85dc6380b057e5b99c9bebfd9147dea76b4e45c4b90280a6fd04L35-R36) [[6]](diffhunk://#diff-c94738844369af11fca2d510cc4e7be994a2d90ea7244dd558e9594e372eb7f4L45-R45)

### Improvements to Testing Framework:
* Added a comprehensive test for ownership handling in `ingestor/storage/crds_test.go` to ensure correct behavior when claiming and checking resource ownership.
* Removed tests related to leader election from `ingestor/storage/kql_functions_test.go` as they are no longer relevant.

These changes collectively enhance the robustness of resource management in the `ingestor` package by ensuring proper ownership handling and simplifying the codebase.

### DIFF
--- a/ingestor/storage/crds.go
+++ b/ingestor/storage/crds.go
@@ -5,10 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"time"
 
 	adxmonv1 "github.com/Azure/adx-mon/api/v1"
 	"github.com/Azure/adx-mon/pkg/logger"
-	"github.com/Azure/adx-mon/pkg/scheduler"
+	"github.com/google/uuid"
 	meta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -36,30 +37,18 @@ type CRDHandler interface {
 }
 
 type crdHandler struct {
-	Client  client.Client
-	Elector scheduler.Elector
+	Client       client.Client
+	ControllerId string
 }
 
-func NewCRDHandler(client client.Client, elector scheduler.Elector) CRDHandler {
+func NewCRDHandler(client client.Client) CRDHandler {
 	return &crdHandler{
-		Client:  client,
-		Elector: elector,
+		Client:       client,
+		ControllerId: "controller-" + uuid.NewString(),
 	}
 }
 
 func (c *crdHandler) List(ctx context.Context, list client.ObjectList, filters ...ListFilterFunc) error {
-	if c.Elector != nil && !c.Elector.IsLeader() {
-		return nil
-	}
-
-	// TODO (jesthom) Method is invoked for each database for each task, we probably want to
-	// switch to some sort of shared cache.
-	// controller-runtime implements such a caching mechanism
-	//
-	// mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{})
-	// then whenever you retrieve a client, you intherit the caching mechanism
-	// client := mgr.GetClient()
-
 	if c.Client == nil {
 		return errors.New("no client provided")
 	}
@@ -74,11 +63,25 @@ func (c *crdHandler) List(ctx context.Context, list client.ObjectList, filters .
 		if !ok {
 			return nil
 		}
+
+		// Apply any filters before checking ownership. We want to ignore ownership
+		// if we're otherwise not interested in this instance. For example, if this
+		// instance has already been reconciled and is otherwise up-to-date, no need
+		// to try and establish ownership. This will prevent us from thrashing by
+		// constantly updating the resource last-updated annotation.
 		for _, filter := range filters {
 			if filter(obj) {
 				return nil
 			}
 		}
+
+		if !CheckOwnership(ctx, obj, c.Client, c.ControllerId) {
+			if logger.IsDebug() {
+				logger.Debugf("Skipping %s/%s, not owned by this controller", obj.GetNamespace(), obj.GetName())
+			}
+			return nil
+		}
+
 		filtered = append(filtered, obj)
 		return nil
 	})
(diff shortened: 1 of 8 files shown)


ingestor/storage/crds.go
```diff
@@ -164,3 +167,61 @@ func ConvertToTypedList(objects []client.Object, list client.ObjectList) error {
...
+
+	// If unowned, claim it
+	if !owned {
+		if logger.IsDebug() {
+			logger.Debugf("Claiming unowned resource %s/%s", obj.GetNamespace(), obj.GetName())
+		}
+		annotations[adxmonv1.ControllerOwnerKey] = ownerId
+		annotations[adxmonv1.LastUpdatedKey] = metav1.Now().Format(time.RFC3339)
+		obj.SetAnnotations(annotations)
+		if err := ctrlClient.Update(ctx, obj); err == nil {
```
jwilder, line 190: Does this handle the case where another controller tried to update at the same time?  i.e. does this use optimisitic concurrency control and return an error if the obj was updated?

ingestor/storage/crds.go
```diff
@@ -164,3 +167,61 @@ func ConvertToTypedList(objects []client.Object, list client.ObjectList) error {
...
+			if logger.IsDebug() {
+				logger.Debugf("Claimed ownership of %s/%s: %v", obj.GetNamespace(), obj.GetName(), err)
+			}
+			owner = ownerId
+		}
```
jwilder, line 195: Handle err != nil case?

ingestor/storage/crds.go
```diff
@@ -164,3 +167,61 @@ func ConvertToTypedList(objects []client.Object, list client.ObjectList) error {
...
+	}
```
jwilder, line 196: Missing return?

ingestor/storage/crds.go
```diff
@@ -164,3 +167,61 @@ func ConvertToTypedList(objects []client.Object, list client.ObjectList) error {
...
+				annotations[adxmonv1.ControllerOwnerKey] = ownerId
+				annotations[adxmonv1.LastUpdatedKey] = metav1.Now().Format(time.RFC3339)
+				obj.SetAnnotations(annotations)
+				if err := ctrlClient.Update(ctx, obj); err == nil {
+					if logger.IsDebug() {
+						logger.Debugf("Claimed ownership of %s/%s: %v", obj.GetNamespace(), obj.GetName(), err)
+					}
+				}
+			} else {
+				return false // Skip, not stale
```
jwilder, line 215: What happens is the timestamp failed to parse?  Perhaps it got updated by somethign else to a non parseable value?

ingestor/storage/crds.go
```diff
@@ -164,3 +167,61 @@ func ConvertToTypedList(objects []client.Object, list client.ObjectList) error {
...
+		if hasLastUpdated {
+			lastTime, err := time.Parse(time.RFC3339, lastUpdated)
+			if err == nil && time.Since(lastTime) > 5*time.Minute { // Stale after 5 minutes
+				if logger.IsDebug() {
+					logger.Debugf("Taking over stale resource %s/%s from %s", obj.GetNamespace(), obj.GetName(), owner)
+				}
+				annotations[adxmonv1.ControllerOwnerKey] = ownerId
+				annotations[adxmonv1.LastUpdatedKey] = metav1.Now().Format(time.RFC3339)
+				obj.SetAnnotations(annotations)
+				if err := ctrlClient.Update(ctx, obj); err == nil {
```
jwilder, line 209: err != nil case is skipped.

ingestor/storage/crds.go
```diff
@@ -164,3 +167,61 @@ func ConvertToTypedList(objects []client.Object, list client.ObjectList) error {
...
+				return false // Skip, not stale
+			}
+		} else {
+			return false // Skip, owned but no timestamp
+		}
+	}
+
+	// We own this instance, set the last-updated annotation to keep ownership active
+	annotations[adxmonv1.LastUpdatedKey] = metav1.Now().Format(time.RFC3339)
+	obj.SetAnnotations(annotations)
```
jwilder, line 224: Don't we need to all ctrl.Update(obj) here?

jwilder (changes requested):

jwilder:
One general comment is that this introduces another "partitioning" concept and implementation into the code.  The other code that has these concepts in in `cluster` package and is based on rendezvous hashing for distribution.  This idea for isntances taking ownership of a CRD could leverage that existing impl as well.

ingestor/storage/crds.go
jessejlt: CRDs have an immutable property `ResourceVersion` added by Kubernetes that acts as a sort of etag in this case. In this case, the first controller / ingestor instance that successfully updates the CRD, the ResourceVersion will change such that the next controller that attempts the Update will be rejected because the ResourceVersion has been modified.

ingestor/storage/crds.go
jessejlt: Errors here are expected due to conflicting ResourceVersions. As two ingestor instances try to take ownership of the same object, one will become the owner and the other will generate an error, which can be ignored.

ingestor/storage/crds.go
jessejlt: Since the owner isn't set, the next block will not execute

ingestor/storage/crds.go
jessejlt: Perhaps we should consider this a failed ownership claim and go ahead and claim ownership. Does that seem like a reasonable approach?

ingestor/storage/crds.go
jessejlt: At this point we own the object, we update our timestamp annotation and the actual persistence of the object will occur after we perform the work described in the CRD.

ingestor/storage/crds.go
jwilder: Seems alright.  We should definitely log the error though so we can discover that it's stuck in an unclaimed state indefinitely.

ingestor/storage/crds.go
jwilder: I'm just concerned that we might be swallowing an different error that we should log.  Would it make sense to explicitly check for the conflict error and ignore, but log the other cases that we would not expect?

ingestor/storage/crds.go
jwilder: I see it's owned, but the last updated key would be stale, right?

PR #620, opened 2025-03-11
Add drop-prefix list argument to ingestor

mkeesey:
This argument provides a prefix for transfer filenames that will be rejected on transfer. This provides a break-glass mechanism for rejecting certain logs or metrics from being processed by a given Ingestor cluster.

### DIFF
--- a/cmd/ingestor/main.go
+++ b/cmd/ingestor/main.go
@@ -62,6 +62,7 @@ func main() {
 			&cli.Int64Flag{Name: "max-segment-count", Usage: "Maximum segment files allowed before signaling back-pressure", Value: 10000},
 			&cli.DurationFlag{Name: "max-transfer-age", Usage: "Maximum segment age of a segment before direct kusto upload", Value: 90 * time.Second},
 			&cli.DurationFlag{Name: "max-segment-age", Usage: "Maximum segment age", Value: 5 * time.Minute},
+			&cli.StringSliceFlag{Name: "drop-prefix", Usage: "Drop transfers that match the file prefix. Transfer filenames are in the form of DestinationDB_Table_..."},
 			&cli.BoolFlag{Name: "enable-wal-fsync", Usage: "Enable WAL fsync", Value: false},
 			&cli.IntFlag{Name: "max-transfer-concurrency", Usage: "Maximum transfer requests in flight", Value: 50},
 			&cli.IntFlag{Name: "partition-size", Usage: "Maximum number of nodes in a partition", Value: 25},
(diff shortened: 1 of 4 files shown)


cmd/ingestor/main.go
```diff
@@ -62,6 +62,7 @@ func main() {
 			&cli.Int64Flag{Name: "max-segment-count", Usage: "Maximum segment files allowed before signaling back-pressure", Value: 10000},
 			&cli.DurationFlag{Name: "max-transfer-age", Usage: "Maximum segment age of a segment before direct kusto upload", Value: 90 * time.Second},
 			&cli.DurationFlag{Name: "max-segment-age", Usage: "Maximum segment age", Value: 5 * time.Minute},
+			&cli.StringSliceFlag{Name: "drop-transfer-prefix", Usage: "Drop transfers that match the file prefix. Transfer filenames are in the form of DestinationDB_Table_..."},
```
jwilder, line 65: What do you think about just `drop-prefix`?  Wondering if `transfer` is necessary or too specific.

ingestor/service.go
```diff
@@ -372,6 +377,15 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
 		}
 	}()
 
+	for _, prefix := range s.dropFilePrefixes {
+		if strings.HasPrefix(filename, prefix) {
+			logger.Errorf("Transfer request rejected due to prefix %q: %q", prefix, filename)
```
jwilder, line 382: I think this could get very noisy in the logs.

ingestor/service.go
```diff
@@ -372,6 +377,15 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
...
+			m.WithLabelValues(strconv.Itoa(http.StatusBadRequest)).Inc()
+			http.Error(w, "filename is configured to be dropped", http.StatusBadRequest)
```
jwilder, line 384: Wondering if we should just silently drop this instead.  e.g. just return a 200 that we accepted and move on.

ingestor/service.go
mkeesey: Very true. Wanted some way to easily track this, but think tracing is probably a better mechanism in the future.

ingestor/service.go
mkeesey: I think we should probably do that. Drop implies we're accepting it but not forwarding it on, but not returning an error either.

ingestor/service.go
```diff
@@ -372,6 +378,15 @@ func (s *Service) HandleTransfer(w http.ResponseWriter, r *http.Request) {
 		}
 	}()
 
+	for _, prefix := range s.dropFilePrefixes {
+		if strings.HasPrefix(filename, prefix) {
+			io.Copy(io.Discard, r.Body)
```
jwilder, line 383: Do we have a metric that we can incremented to track we dropped them?

jwilder (approved):

ingestor/service.go
mkeesey: We now have one per prefix

PR #629, opened 2025-03-13
Record table size data regularly

jwilder:
This will be useful to understand large table growth and how much each table contributes to hot cache/cluster size.

This runs `.show tables details` every minute and stores the output in `AdxmonIngestorTableDetails` so we can have this data tracked over time.


jessejlt (approved):
Looks like something that could turn into a SummaryRule once they're ready

PR #637, opened 2025-03-21
Encapsulate Log object with freeze mechanism

mkeesey:
The main encapsulation changes are in collector/logs/types/logs.go. Prior to going forward with other encapsulation changes which involve a decent number of updates throughout the codebase, I want to float this in front of folks.

The lifecycle of log objects is:

Get created/reset
Modified in a chain of a source, then a succession of filters. This process is intended to be single-threaded or sent between stages through a channel to synchronize the changes and to pass ownership onwards.
Get sent out in parallel to one or multiple outputs. At this point, the Log object should be considered immutable. Modifications are possible with copies, or ideally in-line as the objects are being serialized by being selective around what fields are written, what logs are skipped, etc.
Go does not have an immutability construct that is built around this during compilation. This path is also part of the hot loop portion of the system, so forcing copies in every case is likely to greatly increase cpu and memory usage to support this defensive case. Instead, this proposes to use runtime checks with a very low overhead check during each modification operation to ensure that the object has not been marked as "frozen" which is intended to be done prior to sending to the outputs in parallel. In microbenchmarks this appears to have a negligible impact. The panic response is likely to be hit in testing with any usage of an incorrectly written output or exporter for logs which helps avoid invalid Log object usage even in development.

Encapsulation may have further benefits in the future - we may change the underlying datastructure for Body, Attributes, and Resources away from map[string]any to be even more cache and allocation friendly.

Thus far just timestamps have been encapsulated, but assuming this approach is considered reasonable the rest of the fields within the struct will be encapsulated throughout the codebase as well.


collector/logs/transforms/plugin/test-plugin/src/github.com/Azure/testplugin/vendor/github.com/Azure/adx-mon/collector/logs/types/logs.go
```diff
@@ -17,44 +19,232 @@ type Log struct {
...
-		copy.Body[k] = v
+	for k, v := range l.body {
+		copy.body[k] = v
 	}
-	for k, v := range l.Resource {
-		copy.Resource[k] = v
+	for k, v := range l.resource {
+		copy.resource[k] = v
 	}
 	return copy
```
jessejlt, line 88: So the implication of copying a frozen log is that the copy is thawed while the primary remains frozen, correct?

collector/logs/types/logs.go
```diff
@@ -17,44 +30,238 @@ type Log struct {
...
+	l.checkWrite()
+	l.resource[key] = value
+}
+
+func (l *Log) checkWrite() {
+	if atomic.LoadUint32(&l.frozen) == 1 {
+		if assertionsEnabled() {
+			panic("write to frozen log - not safe for updates")
+		} else {
+			logger.Error("write to frozen log - not safe for updates")
```
jessejlt, line 134: Receiving this error is something we want to address right away, but I'm wondering if we have sufficient information from this log statement to effectively do so. Does it make sense to invoke `runtime.Stack` so that we can better triage and repair the offending write?

jessejlt (approved):
Couple questions. Looks great. I like this approach!

collector/logs/types/logs.go
jwilder: Is there much overhead to calling assertionsEnabled() which looks up os.Environ?  Wondering if checking assertionsEnabled in an `init()` func an setting a package bool var would alleviate that.

collector/logs/transforms/plugin/test-plugin/src/github.com/Azure/testplugin/vendor/github.com/Azure/adx-mon/collector/logs/types/logs.go
mkeesey: No thawing or freezing - just a shallow copy into a separate map that is safe to modify the structure of. Updating the comment to describe a little bit more.

collector/logs/types/logs.go
mkeesey: Originally, I implemented with an init function which did seem to add quite a lot of impact, but now I cannot reproduce that. Swapping back to the init function version to make this simpler.

Good idea with capturing the stack

PR #659, opened 2025-04-07
Fix lifecycle for logs services

mkeesey:
This fixes a few issues with the lifecycle management of logs services.

1. We were not assigning transforms to the service correctly, preventing them from being started and stopped.
2. We were incorrectly closing the source and transforms at the end of the `Open` method with defers, instead of closing them in the `Close` method.
3. The Otel logs service now uses the same workflow as other logs services, ensuring it gets started and stopped correctly.

### DIFF
--- a/collector/logs/service_test.go
+++ b/collector/logs/service_test.go
@@ -2,6 +2,7 @@ package logs_test
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
@@ -9,7 +10,14 @@ import (
 	"github.com/Azure/adx-mon/collector/logs/engine"
 	"github.com/Azure/adx-mon/collector/logs/sinks"
 	"github.com/Azure/adx-mon/collector/logs/sources"
+	"github.com/Azure/adx-mon/collector/logs/types"
+	"github.com/Azure/adx-mon/metrics"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/stretchr/testify/require"
 )
+
+const sourceName = "ConstSource"
 
 func BenchmarkPipeline(b *testing.B) {
 	for i := 0; i < b.N; i++ {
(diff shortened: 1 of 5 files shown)


collector/logs/service_test.go
```diff
@@ -43,3 +45,135 @@ func TestPipeline(t *testing.T) {
...
+
+	t.cancel()
+	return nil
+}
+
+func (t *transformerpanicifclosed) Transform(ctx context.Context, batch *types.LogBatch) (*types.LogBatch, error) {
+	// Panic if the context is already cancelled
+	select {
+	case <-t.ctx.Done():
+		panic("context already cancelled")
```
jwilder, line 168: Shouldn't this just return an error vs panic?

collector/logs/service_test.go
```diff
@@ -43,3 +45,135 @@ func TestPipeline(t *testing.T) {
...
+
+	t.ctx, t.cancel = context.WithCancel(context.Background())
+	return nil
+}
+
+func (t *transformerpanicifclosed) Close() error {
+	// Panic if the context is already cancelled
+	select {
+	case <-t.ctx.Done():
+		panic("context already cancelled")
```
jwilder, line 155: Return error vs panic?

collector/logs/service_test.go
```diff
@@ -43,3 +45,135 @@ func TestPipeline(t *testing.T) {
...
+		worker.Run()
+	}()
+	return nil
+}
+
+func (s *sourceSendOnClose) Close() error {
+	// Panic if the context is already cancelled
+	select {
+	case <-s.ctx.Done():
+		panic("context already cancelled")
```
jwilder, line 112: Return error vs panic?

jwilder (changes requested):

jessejlt (approved):

collector/logs/service_test.go
mkeesey: I think we want panics for this test-only impl to quickly assert.

collector/logs/service_test.go
mkeesey: I think we want panics for this test-only impl to quickly assert.

collector/logs/service_test.go
mkeesey: I think we want panics for this test-only impl to quickly assert.

jwilder: Ah. Missed this was a test impl.

collector/logs/service_test.go
jwilder: Make sense

PR #667, opened 2025-04-12
Logging service and workers accept multiple sinks

mkeesey:
The primary update of this PR is to update the logging worker to send log batches in parallel to all of the configured destination sinks. In addition, the service object that manages the lifecycle of starting and stopping these sinks now accepts a slice of sinks instead of a single sink.


collector/logs/engine/worker.go
```diff
@@ -52,15 +58,23 @@ func (w *worker) processBatch(ctx context.Context, batch *types.LogBatch) {
...
-		// Nack batch?
-		return
+
+	var wg sync.WaitGroup
+	wg.Add(len(w.Sinks))
+	for _, sink := range w.Sinks {
+		go func(sink types.Sink) {
+			defer wg.Done()
+
+			err = sink.Send(ctx, batch)
```
jessejlt, line 68:
```suggestion
			err := sink.Send(ctx, batch)
```
I think this is a race as each sink can modify the err variable, which appears to be declared outside the scope of the go-routines.

jessejlt (changes requested):

jessejlt (approved):

PR #669, opened 2025-04-16
Add support for gzipping segment transfers

jwilder:
This adds an option to gzip segments during transfer.  The intent is that this would be enabled by default for collector to ingestor but probably off for ingestor toingestor as they are on the same vnet.


jessejlt (approved):
Fantastic!

matucker-msft (approved):
Beautiful <3

PR #670, opened 2025-04-16
Design doc for proposed adx-operator

jessejlt:


docs/designs/operator.md
```diff
@@ -0,0 +1,201 @@
...
+  - Watch the manifests for adx-mon components.
+  - If a user manually changes a managed resource (e.g., increases ingestor replicas), detect the change and update the operator CRD's status/spec to reflect the new state.
+  - Ensure the operator's desired state is always in sync with the actual cluster state, supporting both declarative and imperative workflows.
+
+- **Incremental, Granular Reconciliation:**  
+  - The operator proceeds in highly granular steps, maintaining subresource conditions for each phase (e.g., ADX cluster ready, database ready, ingestor ready, collector ready).
+  - Each phase is only attempted after all dependencies are successfully reconciled.
+  - Status and conditions are updated at each step to reflect progress and issues.
+
+- **Resource Cleanup on Deletion:**
```
jwilder, line 34: Not sure if this should be the default behavior.  I like the idea that ASO has where it can detach from the azure resources.  Deleting these resources should be more explict IMO like marking a resource to be deleted vs just deleting the resource.  The latter could easily happen from bugs or operator error.

docs/designs/operator.md
```diff
@@ -0,0 +1,201 @@
...
+- **adx-mon Component Management:**  
+  - Generate and apply manifests for adx-mon components (collector, ingestor, alerter) as Kubernetes resources (e.g., StatefulSets, Deployments, Services).
+  - Support default images for each component, with the ability to override via CRD spec.
+  - Allow configuration of the number of ingestor instances (replicas).
+  - Support further customization of component manifests via CRD fields.
+  - Support deployment of only a subset of components (e.g., only collectors, or only ingestors) to enable federated/multi-cluster topologies.
+
+- **Reconciliation and Drift Detection:**  
+  - Watch the manifests for adx-mon components.
+  - If a user manually changes a managed resource (e.g., increases ingestor replicas), detect the change and update the operator CRD's status/spec to reflect the new state.
```
jwilder, line 26: re: scaling ingestor... that is something that the operator should be able to handle.  Just thinking out loud, but we might want a `manual` mode as well as an `auto-scaling` mode to allow a user to control the scaling if they want.

docs/designs/operator.md
```diff
@@ -0,0 +1,201 @@
...
+    - Deploys ingestor components (and optionally ADX resources).
+    - Receives events from multiple remote collector clusters.
+  - **Collector Clusters:**  
+    - Deploys only collector components.
+    - Forwards events to a central ingestor cluster.
+    - Requires configuration of ingestor endpoints and authentication.
+
+- The CRD spec determines the deployment mode:
+  - If only `collector` is specified, the operator assumes this is a collector-only cluster and requires `ingestor` connection details.
+  - If `ingestor` and `adx` are specified, the operator will bootstrap the full stack and expose endpoints for remote collectors.
```
jwilder, line 52: I was thinking we'd have sections for each of the components and depending on what is defined, that part gets created.  Maybe each sections is separate CRD?

docs/designs/operator.md
```diff
@@ -0,0 +1,201 @@
...
+  - Cluster connection info, including endpoint and MSI client-id if using managed identity.
+  - Optional: advanced overrides for component manifests.
+  - **Federation fields:**  
+    - For collector-only clusters: specify `ingestor` connection details (endpoint, authentication).
+    - For ingestor clusters: expose endpoints for remote collectors.
+
+- **Status fields:**  
+  - Reflect the current state and subresource conditions (e.g., ADX cluster ready, database ready, ingestor ready, collector ready).
+
+### Example CRD
```
jwilder, line 83: I like the feel of this.  Very simple and straight-forward.    We should explore what more complicated setup options we might need to support and see if it would make this CRD too complicated.  For example, defining singleton collectors or more complicated collector configmaps.

docs/designs/operator.md
```diff
@@ -0,0 +1,201 @@
...
+     - Create collector DaemonSet (parameterized with ingestor endpoints).
+     - Wait for collector to be ready.
+     - Create alerter if specified.
+   - Each step is only attempted after all dependencies are ready.
+
+2. **Watch Managed Resources:**  
+   - Monitor StatefulSets/Deployments for adx-mon components.
+   - If a managed resource is changed outside the operator (e.g., replica count updated), update the Operator CRD to reflect the new state.
+
+3. **Sync Desired and Actual State:**
```
jwilder, line 181: #3 conflicts with #2.  For example, say the SKU for an ADX cluster is changed on the portal.  Should the CRD get updated to reflect that or should it update ADX back to the SKU defined in the CRD?

For ADX, I kind of lean towards using the CRDs for creating the initial resource or adopting an existing one to learn about endpoints, but leave the internal state reconcilation alone.  So in the prior case, if the ADX cluster exists, just don't pay attention to replica count or SKU size, etc..

For ingestor, collector, etc.. that might not make as much sense though.  For example, updating the ingestor image version should probably sync from CRD to actual state and revert if the actual state was updated.

docs/designs/operator.md
```diff
@@ -0,0 +1,201 @@
...
+3. **Sync Desired and Actual State:**  
+   - If the Operator CRD is updated, reconcile the manifests and Azure resources as needed.
+   - If the actual state drifts from the desired state, update the CRD status/spec accordingly.
+
+---
+
+## Implementation Notes
+
+- The operator should use controller-runtime's watches and informers to efficiently detect changes to both its own CRD and the managed resources.
+- ASO CRDs are the source of truth for Azure resources; the operator should not attempt to manage Azure resources directly.
```
jwilder, line 190: Don't think I'm comfortable with this just yet.

docs/designs/operator.md
jessejlt: ASO is left as an implementation option, but the current default implementation is to maintain ADX clusters using the azure-SDK directly.

docs/designs/operator.md
```diff
@@ -0,0 +1,201 @@
+# adx-mon Operator Design
+
+## Overview
+
+The `adx-mon-operator` is a Kubernetes operator responsible for managing the lifecycle of an adx-mon cluster, including its core components (collector, ingestor, alerter) and required Azure infrastructure. The operator leverages [Azure Service Operator (ASO)](https://azure.github.io/azure-service-operator/) to declaratively manage Azure resources such as ADX (Azure Data Explorer) clusters, databases, and other dependencies.
```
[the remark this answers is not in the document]
jessejlt: ASO is left as an implementation option, but the current default implementation is to maintain ADX clusters using the azure-SDK directly.

docs/designs/operator.md
mkeesey: I may misunderstand the structure, but I wouldn't think that the separate CRDs for each component would need to know about each other, but instead they'd reconcile independently like many k8s resources (e.g. service objects point at pods based on labels, but reconcile completely separately). Many operators in the community use different CRDs for different resources (e.g. prometheus operator, etc).

I think this will make each reconciler within the operator simpler in the long term since it is concerned with reconciling a single CRD kind rather than a big block of tons of resources. I think it makes the configuration more orthogonal and consistent as well - if you want multiple sets of collectors in a cluster (say for different node pools) you configure the second set of collectors just like the first ones.

docs/designs/operator.md
```diff
@@ -0,0 +1,201 @@
...
+
+The operator aims to provide a simple, production-ready bootstrap experience for adx-mon clusters, while supporting advanced customization for complex and federated deployments.
+
+---
+
+## Responsibilities
+
+- **Azure Infrastructure Management:**
```
vpatelsj, line 13: I am curious how managing azure infra using operators pans out because there will be two source of truths - one state in kubernetes and one in azure. The straight greenfield case of creating azure resources from scratch works fine in my head right now but the day 2 operations and reconciliation complications when state changes outside of operator, corner cases where azure apis return errors makes things hairy imo. Just to clarify, I am not objecting. I am actually open to trying this out knowing that we can always end up going back to a bootstrap script just for the infrastructure part.

docs/designs/operator.md
vpatelsj: I like the idea of separate CRDs for each component. How do we ensure all component crds are updated at the same time? Technically by having independent CRDs there is a chance there is a mismatch in version cadence and might lead to unexpected behavior which i guess becomes easier to spot when you have one giant CRD.

docs/designs/operator.md
jwilder: > This is a good example. To slightly flip the example, what would you expect as a CRD author if you initially want SKU A, but later manually change the SKU to B, and even later down the road update the CRD to SKU C?

So, if CRD expresses the desired state and azure is the actual state and SKU A is defined initially... the change to SKU B would get reverted by the operator.  Alternatively, it could fail to reconcile and an human operator would need to reconcile the CRD with the actual state as this is not an expected change and we don't want to cause impact to the cluster.

> I think we should remove https://github.com/Azure/adx-mon/pull/2 and the same block below though that describe the CRD spec getting updated based on changes to the actual resources though. We want the CRD to manage it or not to manage it, not have it both ways.

Agree.  I think only Status/conditions should get updated from actual state.

docs/designs/operator.md
```diff
@@ -0,0 +1,592 @@
...
+
+The operator supports both fully specified and zero-config deployments:
+
+1. **Zero-Config Mode:**
+   - Uses Azure IMDS to automatically detect:
+     - Region
+     - Subscription ID
+     - Resource Group
+     - AKS cluster name (for naming)
+   - Automatically selects optimal SKU based on regional availability
```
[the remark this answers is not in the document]
jwilder: I think this might be impled for ADX sku selection where we want an storage optimized L series right now.

PR #671, opened 2025-04-16
improvement: add timestamp to shutdown annotation for better tracking

vpatelsj:


matucker-msft (approved):
Thanks for switching to patch!

PR #678, opened 2025-04-25
Finish connecting the fluent-output exporter

mkeesey:
This pull request introduces support for Fluent Forward log exporters in the configuration system, including validation, integration, and testing. The changes primarily focus on adding a new log exporter type, ensuring its configuration is valid, and integrating it into the existing logging pipeline.

This adds new configuration properties for this new exporter and configures our log sources to be able to use them. As part of this, I moved the configuration of OtelLog (the http endpoint) into main.go alongside our other configuration blocks so it can be configured with the config object as well.


cmd/collector/config/export.go
```diff
@@ -176,3 +218,46 @@ func getLabelMappings(mappings []LabelMatcher) (map[*regexp.Regexp]*regexp.Regex
...
+	}
+	if f.Destination == "" {
+		return fmt.Errorf("destination must be set")
+	}
+
+	if strings.HasPrefix(f.Destination, "unix://") {
+		socketPath := strings.TrimPrefix(f.Destination, "unix://")
+		if len(socketPath) == 0 {
+			return fmt.Errorf("invalid destination %s: unix socket path is empty", f.Destination)
+		}
```
jessejlt, line 240: Should we verify the path exists and we have permissions to read from it?

```suggestion
		}
		fileInfo, err := os.Stat(path)
	        if err != nil {
		    if os.IsNotExist(err) {
			return fmt.Errorf("path does not exist: %s", path)
			}
	
	}
```

collector/export/log_fluent.go
```diff
@@ -86,14 +104,7 @@ func (e *LogToFluentExporter) dial(ctx context.Context) (net.Conn, error) {
...
 
-	// Check if destination starts with "unix://" for Unix domain socket
-	if strings.HasPrefix(e.destination, "unix://") {
-		socketPath := strings.TrimPrefix(e.destination, "unix://")
-		return d.DialContext(ctx, "unix", socketPath)
-	}
-
-	// Default to TCP
-	return d.DialContext(ctx, "tcp", e.destination)
+	return d.DialContext(ctx, e.destinationNetwork, e.destination)
```
jessejlt, line 107: If the receiver is down, won't this cause `collector` instances to crash-loop until the receiver is back up? Maybe we should attempt to dial at each scrape interval instead?

jessejlt (approved):

collector/export/log_fluent.go
mkeesey: It will not cause crashes. Each sink independently sends and tracks their own failures, but even if this sender is stuck not being able to forward we will continue sending to ingestor. https://github.com/Azure/adx-mon/blob/main/collector/logs/engine/worker.go#L75

cmd/collector/config/export.go
mkeesey: I don't think so. We don't check file paths (or socket destinations) in our config validation right now. I think it's good that we don't so that we can more easily create a validate subcommand to check that a config is well-formed without needing to be on an actual host with all of the available paths, etc.

PR #685, opened 2025-04-29
Add initial design documentation for adx-mon operator

jessejlt:


docs/designs/operator.md
```diff
@@ -0,0 +1,348 @@
...
+
+This example demonstrates all available configuration options for each CRD:
+
+```yaml
+apiVersion: adx-mon.azure.com/v1
+kind: ADXCluster
+metadata:
+  name: prod-adx-cluster
+spec:
+  clusters:
```
jessejlt, line 241: I think I'm going to change this to be one cluster per CRD. Having a list of clusters in the CRD makes the status object a bit confusing and difficult to reason about.

docs/designs/operator.md
```diff
@@ -0,0 +1,348 @@
...
+---
+
+apiVersion: adx-mon.azure.com/v1
+kind: Ingestor
+metadata:
+  name: prod-ingestor
+spec:
+  image: "ghcr.io/azure/adx-mon/ingestor:v1.0.0"
+  replicas: 3
+  adxClusterRef: prod-adx-cluster
```
jessejlt, line 268: This will need to be a list.

docs/designs/operator.md
```diff
@@ -0,0 +1,348 @@
...
+metadata:
+  name: prod-collector
+spec:
+  image: "ghcr.io/azure/adx-mon/collector:v1.0.0"
+  ingestorEndpoint: "http://prod-ingestor.monitoring.svc.cluster.local:8080"
+  ingestorAuth:
+    type: "token"
+    tokenSecretRef:
+      name: "ingestor-auth"
+      key: "token"
```
jessejlt, line 283: Auth is not yet well defined. I'll retrieve the Endpoint from the Ingestor spec. The idea being that if running out-of-cluster, at some point in a future iteration, the operator will generate the connection details and make them available on its own CRD.

docs/designs/operator.md
jessejlt: Also will add Endpoint details for Collector to consume

```go
        // +kubebuilder:validation:Optional
	// Endpoint is the endpoint to use for the ingestor. If running in a cluster, this should be the service name
	// otherwise the operator will generate an endpoint.
	Endpoint string `json:"endpoint,omitempty"`

	// +kubebuilder:validation:Optional
	// +kubebuilder:default=false
	// ExposeExternally is a flag to indicate if the ingestor should be exposed externally as reflected in the Endpoint.
	ExposeExternally bool `json:"exposeExternally,omitempty"`
```

docs/designs/operator.md
```diff
@@ -0,0 +1,348 @@
...
+
+---
+
+apiVersion: adx-mon.azure.com/v1
+kind: Alerter
+metadata:
+  name: prod-alerter
+spec:
+  image: "ghcr.io/azure/adx-mon/alerter:v1.0.0"
+```
```
jessejlt, line 293: Will add `NotificationEndpoint` and ADX cluster references for endpoints.

docs/designs/operator.md
```diff
@@ -0,0 +1,348 @@
...
+     - Create ingestor StatefulSet (parameterized with ADX connection strings).
+     - Wait for ingestor to be ready.
+     - Create collector DaemonSet (parameterized with ingestor endpoints).
+     - Wait for collector to be ready.
+     - Create alerter if specified.
+   - Each step is only attempted after all dependencies are ready.
+
+2. **Watch Managed Resources:**  
+   - Monitor StatefulSets/Deployments for adx-mon components.
+   - If a managed resource is changed outside the operator (e.g., replica count updated), update the managed resource to match the desired state.
```
mkeesey, line 321: Think we talked about removing this part - maybe we can reword that it updates the status? We want the CRD to own the number of replica counts and other similar fields and reconcile the owned resources to those values. We don't want to have a two-way sync.

docs/designs/operator.md
```diff
@@ -0,0 +1,348 @@
...
+
+3. **Sync Desired and Actual State:**  
+   - If the Operator CRD is updated, reconcile the manifests and Azure resources as needed.
+   - If the actual state drifts from the desired state, update the actual state to match the desired state.
```
mkeesey, line 325: Same as above comment

docs/designs/operator.md
jessejlt: What you describe is the intent. Let me see if I can reword this section a little.

jessejlt: Updated the language.

mkeesey (approved):

PR #692, opened 2025-05-02
Track which Kusto cluster Functions were applied to

jessejlt:
Ensure KQL functions are synchronized correctly even if the target Kusto endpoint changes.

### DIFF
--- a/api/v1/function_types.go
+++ b/api/v1/function_types.go
@@ -20,27 +20,29 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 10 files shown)


api/v1/function_types.go
```diff
@@ -20,27 +20,29 @@ import (
...
 // AllDatabases is a special value that indicates all databases
 // +k8s:deepcopy-gen=false
 const AllDatabases = "*"
 
 // FunctionSpec defines the desired state of Function
 type FunctionSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
+	//+kubebuilder:validation:Required
```
jessejlt, line 29: oops, fixing

PR #695, opened 2025-05-03
Proposal: Add cluster federation support to the ADX cluster operator.

jessejlt:


docs/designs/operator.md
```diff
@@ -304,6 +304,153 @@ This configuration demonstrates:
...
+#### For Partition Clusters
+
+```yaml
+spec:
+  role: Partition
+  federation:
+    federatedClusters:
+      - endpoint: "https://federated.kusto.windows.net"
+        heartbeatDatabase: "FleetDiscovery"
+        heartbeatTable: "Heartbeats"
```
mkeesey, line 330: nit: Does it make sense to have this at the top-level and not part of each list element?

docs/designs/operator.md
```diff
@@ -304,6 +304,153 @@ This configuration demonstrates:
...
+```
+- `federatedClusters`: List of federated cluster endpoints and heartbeat config.
+- `partitioning`: Open-ended map/object for partitioning metadata (geo, location, tenant, etc.).
+
+#### For Federated Clusters
+
+```yaml
+spec:
+  role: Federated
+  federation:
```
mkeesey, line 344: Does this also need a reference to the endpoint it applies macro-expands to + a managedIdentity?

mkeesey (approved):

docs/designs/operator.md
jessejlt: I think it's a good idea. I suspect once implementation begins, the API / CRD will change a bit as we get a feel for it, at which point I'll update these docs.

docs/designs/operator.md
jessejlt: The `federated` section is an extension of the existing `ADXCluster` CRD, which already contains an `Endpoint` for the cluster; however, for clarity I've updated the sample spec to show a full configuration.

mkeesey (approved):

PR #696, opened 2025-05-05
Improve integration test reliability

jessejlt:
We sporadically drop our port-forward connectivity between the tests and kustainer. This PR attempts to improve reliability by monitoring the connection and re-establishing when the connection fails.

### DIFF
--- a/pkg/testutils/kustainer/kustainer.go
+++ b/pkg/testutils/kustainer/kustainer.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


pkg/testutils/kustainer/kustainer.go
```diff
@@ -128,6 +135,56 @@ func (c *KustainerContainer) PortForward(ctx context.Context, config *rest.Confi
 	return fmt.Errorf("failed to connect to kustainer after retries: %w", lastErr)
 }
 
+// monitorConnection periodically checks the endpoint and reconnects if needed.
+func (c *KustainerContainer) monitorConnection(ctx context.Context, config *rest.Config, podName string) {
```
mkeesey, line 139: This feels like a lot ... I'm pretty surprised we're dropping simple connections on a single VM. (I do see them in the logs.)

I'd really prefer we turn on more telemetry to find out why this is dropping - maybe Kustainer needs more resources/is crashing/something else? We can beef up the size of the runners https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners but I think it's worth trying to get some telemetry first for why this seems to happen so much.

I can try to help with this in the near future.

PR #725, opened 2025-05-22
docs: add comprehensive documentation for Ingestor WAL storage management and backpressure

jessejlt:


docs/storage.md
```diff
@@ -0,0 +1,187 @@
...
+## Summary Table
+
+| Mechanism         | Purpose                                 | Triggered By                |
+|-------------------|-----------------------------------------|-----------------------------|
+| WAL Write         | Durability, batching                    | Every telemetry write       |
+| Segment Rotation  | Batching, upload/transfer optimization  | Max size/age, periodic task |
+| Peer Transfer     | Coalescing small files                  | Segment < threshold         |
+| Direct Upload     | Timely upload, fallback                 | Segment large/old/transfer fail |
+| Cleanup           | Free disk, prevent storage exhaustion   | After upload/transfer       |
+| Backpressure      | Prevent disk exhaustion, signal overload| On write/transfer if limits exceeded |
```
mkeesey, line 118: Not quite sure about this table - seems to be combining several unrelated things?

docs/storage.md
```diff
@@ -0,0 +1,187 @@
...
+  - Collector will drop any telemetry data it cannot persist to the WAL due to these errors.
+  - There is no upstream backpressure: Collector does not signal Prometheus or other scrapers to slow down; it simply drops data that cannot be written.
+  - No HTTP 429 or similar error is sent to remote write clients; data loss is silent except for logs/metrics.
+- **When limits are not set:**
+  - Collector will continue writing WAL files until the disk is physically full, potentially causing node instability.
+
+## Recommendations
+
+- Always set `max-disk-usage` and/or `max-segment-count` in Collector's configuration to prevent disk exhaustion.
+- Monitor Collector logs and metrics for dropped data due to storage limits.
```
mkeesey, line 164: We should expand what this means - how would one monitor for this?

docs/storage.md
```diff
@@ -0,0 +1,187 @@
...
+| Backpressure      | Propagates to senders/peers                          | No backpressure, silent data loss |
+| Config Defaults   | 10 GiB disk, 10,000 segments if unset                | Unlimited if unset                |
+
+---
+
+## Example: WAL Segment Lifecycle
+
+1. Telemetry arrives and is appended to the WAL segment for its table/schema.
+2. When the segment reaches max size or age, it is rotated.
+3. If the segment is small, it is transferred to a peer for coalescing; otherwise, it is uploaded directly to ADX.
```
mkeesey, line 181: This isn't quite right for collector. We don't transfer between our peers.

docs/storage.md
```diff
@@ -0,0 +1,187 @@
...
+4. After upload/transfer, the segment is deleted from disk.
+5. If disk usage or segment count limits are exceeded, new writes/transfers are rejected until space is freed.
```
mkeesey, line 183: This isn't right - currently we don't reject new writes as far as I know. If it is right, it is in conflict with the table above as we describe the behavior for backpressure.

At least at max segment counts, collector does stop prometheus scrapes but not sure if we consume this elsewhere as well.

docs/storage.md
```diff
@@ -0,0 +1,187 @@
...
+| Config Defaults   | 10 GiB disk, 10,000 segments if unset                | Unlimited if unset                |
+
+---
+
+## Example: WAL Segment Lifecycle
+
+1. Telemetry arrives and is appended to the WAL segment for its table/schema.
+2. When the segment reaches max size or age, it is rotated.
+3. If the segment is small, it is transferred to a peer for coalescing; otherwise, it is uploaded directly to ADX.
+4. After upload/transfer, the segment is deleted from disk.
```
mkeesey, line 182: Just transfer for collector

docs/storage.md
```diff
@@ -0,0 +1,187 @@
...
+### Configurable Limits
+- **Max Disk Usage**: Total bytes allowed for all WAL segments, including:
+  - All closed segments (on disk)
+  - The active (open) segment
+  - Any in-flight (not yet flushed) writes buffered in memory
+- **Max Segment Count**: Maximum number of WAL segments allowed on disk.
+
+### Enforcement
+- On every write, Ingestor checks if adding data would exceed disk usage or segment count limits.
+- **Disk usage enforcement is comprehensive:**
```
mkeesey, line 43: Sort of true. Many of the metrics like current index size are populated async on a timer. There can be gaps currently between when a write has finished (so it's removed from currentInflight) and when the async process updates it.

mkeesey (changes requested):

PR #764, opened 2025-06-05
(proposal) Design document for kusto-to-metrics

jessejlt:
This pull request introduces a comprehensive design for integrating SummaryRule outputs with OTLP metrics export in ADX-Mon. The proposed solution bridges the gap between data aggregation and metrics export by introducing a new `MetricsExporter` CRD, enabling a standardized and declarative approach to creating metrics pipelines. The design prioritizes clean separation of concerns, flexibility, and adherence to Kubernetes best practices.


docs/designs/kusto-to-metrics.md
```diff
@@ -0,0 +1,523 @@
...
+
+**Cons:**
+- Violates single responsibility principle
+- Limits reusability (can't export from multiple SummaryRules to same exporter)
+- Makes SummaryRule CRD more complex
+- Harder to extend with additional export targets
+
+### Approach 2: New MetricsExporter CRD (Recommended)
+
+Create a separate CRD that references SummaryRules and manages metrics export:
```
bmoore-msft, line 69: so my understanding is that we'd need one of these for each OTLP "target"?  So we'd create one for the use case of getting the data into MDM for SLOHub, but since that use case requires uniform data / same schema for all SLIs, they can share a single CRD?

docs/designs/kusto-to-metrics.md
```diff
@@ -0,0 +1,523 @@
...
+The MetricsExporter will integrate with existing ADX-Mon components:
+
+1. **SummaryRuleTask** ([`ingestor/adx/tasks.go`](https://github.com/Azure/adx-mon/blob/main/ingestor/adx/tasks.go)): Continue executing SummaryRules as normal
+2. **OTLP Exporters** ([`collector/export/metric_otlp.go`](https://github.com/Azure/adx-mon/blob/main/collector/export/metric_otlp.go)): Reuse existing exporter infrastructure
+3. **New MetricsExporterTask**: Query SummaryRule output tables, transform to OTLP metrics, and export
+
+### Execution Flow
+
+1. **SummaryRule Execution**: SummaryRuleTask executes KQL queries on schedule, stores results in tables
+2. **Metrics Export Trigger**: MetricsExporterTask runs on its own schedule (independent of SummaryRule cadence)
```
bmoore-msft, line 291: is this one per summary rule or just one for all summary rules?

docs/designs/kusto-to-metrics.md
```diff
@@ -0,0 +1,523 @@
...
+  exporters: ["central-prometheus"]
+  interval: 2m
+  transform:
+    metricNameColumn: "metric_name"
+    valueColumn: "metric_value"
+    timestampColumn: "timestamp"
+    labelColumns: ["Region", "ServiceName", "error_rate"]
+```
+
+## Implementation Timeline
```
bmoore-msft, line 473: Anything specific to sovereign clouds?  or is that all in a different layer?

docs/designs/kusto-to-metrics.md
```diff
@@ -0,0 +1,523 @@
...
+    // Transform defines how to convert query results to metrics
+    Transform TransformConfig `json:"transform"`
+}
+
+type SummaryRuleRef struct {
+    // Name of the SummaryRule in the same namespace
+    Name string `json:"name"`
+    
+    // Database and Table to query for export
+    Database string `json:"database"`
```
matucker-msft, line 92: Why do we need to duplicate the database and table here? wouldnt those always come from the summary rule?

docs/designs/kusto-to-metrics.md
jessejlt: Not that I can think of.

docs/designs/kusto-to-metrics.md
jessejlt: The design has been updated to use a label selector for summary rule targeting, which means authoring one `MetricsExporter` with a `SummaryRule` label selector and for each `SummaryRule` that you want to be included in the `MetricsExporter`, using a common labeling schema.

matucker-msft (approved):

PR #787, opened 2025-06-11
fix: Variable replacements in SummaryRules

jessejlt:
If a SummaryRule's KQL statement is submitted like

```kql
T
| where region == "_region"
```

where the user's intent is to replace `_region` with some value like `eastus`, we would expect the replacement to produce

```kql
T
| where region == "eastus"
```

but we were instead incorrectly double quoting

```kql
T
| where region == "'eastus'"
```

This PR updates the replacement logic to prevent double quoting and the associated tests with the correct expectation.


ingestor/adx/tasks.go
```diff
@@ -449,8 +449,11 @@ func applySubstitutions(body, startTime, endTime string, clusterLabels map[strin
 
 	// Replace cluster label placeholders
 	for k, v := range clusterLabels {
-		placeholder := fmt.Sprintf("%s", k)
-		replacement := fmt.Sprintf("'%s'", v)
+		// Look for quoted placeholders like "key" and replace with properly escaped quoted values
+		placeholder := fmt.Sprintf("\"%s\"", k)
```
jwilder, line 453: Hmm.  String replacement on the query isn't ideal as it can replace the wrong things.  In alerter,  we prepend the query with a new let statement:

`let _region = "foo";` 

We should use the same approach if possible because the string replacement approach will catch incorrect things like a table name "foo_region_bar"

jwilder (changes requested):

ingestor/adx/tasks.go
jessejlt: Management commands do not support preceding statements but must begin with a `.`, for example:

**invalid**

```kql
let _startTime = datetime('2024-01-01');
.set-or-append async MyTable <| 
MyTable | where Timestamp > _startTime
```

However it appears that an embedded `let` statement **is** supported

```kql
.set-or-append async MyTable <| 
let _startTime = datetime('2024-01-01');
MyTable | where Timestamp > _startTime
```

So we _could_ support a SummaryRule like

```yaml
apiVersion: adx-mon.azure.com/v1
kind: SummaryRule
metadata:
  name: hourly-avg-metric
spec:
  database: Metrics
  name: HourlyAvg
  body: |
    SomeMetric
    | where Timestamp between (_startTime .. _endTime)
    | where Region == _region
    | summarize avg(Value) by bin(Timestamp, 1h)
  table: SomeMetricHourlyAvg
  interval: 1h
```

And the task executor could add `let _region="foo"` such that the submitted async operation becomes

```kql
.set-or-append async SomeMetricHourlyAvg <| 
    let _region="eastus"
    SomeMetric
    | where Timestamp between (_startTime .. _endTime)
    | where Region == _region
    | summarize avg(Value) by bin(Timestamp, 1h)
```

Is this what you're advocating for?

matucker-msft: That makes sense to me at least :)

jessejlt: Implemented

matucker-msft: Looks great! I think this is a much safer way than the string replacement!

ingestor/adx/tasks.go
```diff
@@ -443,18 +443,26 @@ func (t *SummaryRuleTask) Run(ctx context.Context) error {
...
+	// Build the wrapped query with let statements, with direct value substitution
+	var letStatements []string
 
-	// Replace cluster label placeholders
+	// Add time parameter definitions with direct datetime substitution
+	letStatements = append(letStatements, fmt.Sprintf("let _startTime=datetime(%s);", startTime))
+	letStatements = append(letStatements, fmt.Sprintf("let _endTime=datetime(%s);", endTime))
+
+	// Add cluster label parameter definitions with direct value substitution
 	for k, v := range clusterLabels {
```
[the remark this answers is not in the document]
jessejlt: Done

matucker-msft (approved):

PR #826, opened 2025-06-26
KQL Query Execution Engine

jessejlt:
#### 3. KQL Query Execution Engine
**Goal**: Implement KQL query execution with time window management
- **Deliverables**:
  - Create ADX client connection management in `adxexporter`
  - Implement time window calculation logic (adapted from SummaryRule patterns)
  - Add KQL query execution with `_startTime`/`_endTime` parameter injection
  - Implement scheduling logic based on `Interval` and last execution tracking
  - Add comprehensive error handling for query failures
- **Testing**: Unit tests with mock ADX responses and integration tests with real ADX
- **Acceptance Criteria**: Can execute KQL queries on schedule with proper time window management


adxexporter/service.go
```diff
@@ -146,3 +156,89 @@ func (r *MetricsExporterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// For now, we'll leave this as a placeholder since we need to
+	// implement the status update mechanism
+	logger.Debugf("Updated last execution time to %s for MetricsExporter %s/%s",
+		endTime.Format(time.RFC3339), me.Namespace, me.Name)
+
+	return nil
+}
+
+// getQueryExecutor gets or creates a QueryExecutor for the specified database
+func (r *MetricsExporterReconciler) getQueryExecutor(database string) (*QueryExecutor, error) {
```
mkeesey, line 217: Creating a new Kusto client for each query will put a lot of load on the token-generation mechanism which makes things like imds grumpy.

Can we reuse `MultiKustoClient` from Alerter and/or maintain a map of db -> kustoclients like we do in there to reuse them?

api/v1/metricsexporter_types.go
```diff
@@ -94,3 +105,99 @@ type MetricsExporterList struct {
...
+		cnd = &metav1.Condition{
+			LastTransitionTime: metav1.Time{Time: clk.Now().Add(-me.Spec.Interval.Duration)},
+		}
+	}
+
+	// Determine if the query should be executed based on several criteria:
+	// 1. The MetricsExporter is being deleted
+	// 2. MetricsExporter has been updated (new generation)
+	// 3. It's time for the next interval execution (based on actual time windows)
+	return me.DeletionTimestamp != nil || // MetricsExporter is being deleted
```
mkeesey, line 168: Do we want to execute if it's being deleted?

mkeesey (changes requested):

adxexporter/service.go
```diff
@@ -146,3 +156,89 @@ func (r *MetricsExporterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	// Calculate the execution window using the CRD method
+	startTime, endTime := me.NextExecutionWindow(clk)
+
+	logger.Infof("Executing MetricsExporter %s/%s for window %s to %s",
+		me.Namespace, me.Name, startTime.Format(time.RFC3339), endTime.Format(time.RFC3339))
+
+	// Execute the KQL query
+	result, err := executor.ExecuteQuery(ctx, me.Spec.Body, startTime, endTime, r.ClusterLabels)
```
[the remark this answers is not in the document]
jessejlt: Timeout added. Kind of a magic number here (5 min), just chose to use the same timeout that `Alerter` has in place.

adxexporter/service.go
jessejlt: I tried extracting `multikustoclient` into `pkg` as a shared implementation between `Alerter` and `adxexporter`. The extraction quickly spiraled as the client package makes use of the `engine` package within `Alerter`. Instead of continuing down the extraction, I decided to just add the database-scoped cache to `adxexporter`.

api/v1/metricsexporter_types.go
jessejlt: Derp, fixed.

mkeesey (approved):

PR #886, opened 2025-08-08
fix(adx): remove unsupported mgmt query parameters; inline validated operation id in getOperation

jessejlt:
It was copilot that wanted to add the query params, turns out they're not supported with management commands, pretty clear in the docs. Grr.


matucker-msft (approved):

api/v1/summaryrule_types.go
```diff
@@ -359,7 +372,7 @@ func (s *SummaryRule) BackfillAsyncOperations(clk clock.Clock) {
 
 	// Start from the last execution time and generate windows forward
 	currentWindowStart := lastExecutionTime.UTC()
-	intervalDuration := s.Spec.Interval.Duration
+	// intervalDuration already captured above
```
matucker-msft, line 375: bad copilot :)

matucker-msft (approved):

PR #907, opened 2025-09-02
feat(collector): Log loss visibility

jessejlt:
## Summary
Phase 0 implementation of log loss visibility as specified in `.github/plans/feature/log-loss-visibility.md`.

## Changes
- Add `log_monitor` config block (allow-list of `database:table` pairs) with validation & warnings.
- Introduce metric `adxmon_collector_monitored_logs_collected_total{database,table}`.
- Extend `StoreSink` to aggregate and emit counts for monitored pairs.
- Wire monitored set through collector main and sinks.
- Add documentation (`docs/monitoring.md`) and nav / index links.
- Basic unit tests for config parsing.

## Notes
- Counting occurs after successful write to storage (slight potential overcount if write fails is accepted per design).
- Counter resets on restart; reconciliation relies on windowed Kusto SummaryRule outputs.

## Follow Ups (Deferred)
- StoreSink test coverage for counting logic.
- Additional metrics for write failures / missing routing info per monitored pair.

Marking as DRAFT for initial review.

### DIFF
--- a/cmd/collector/config/config.go
+++ b/cmd/collector/config/config.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/Azure/adx-mon/collector/logs/sources/tail/sourceparse"
 	"github.com/Azure/adx-mon/collector/logs/transforms"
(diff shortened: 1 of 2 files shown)


cmd/collector/config/config.go
```diff
@@ -111,6 +112,54 @@ type Config struct {
...
+
+type LogMonitor struct {
+	Pairs  []string            `toml:"pairs" comment:"List of database:table pairs to monitor for loss visibility."`
+	parsed map[string]struct{} `toml:"-"`
+}
+
+func (lm *LogMonitor) monitorKey(db, table string) string { return db + "\x00" + table }
+
+// BuildSet parses and validates pairs into the internal set. It is idempotent.
+func (lm *LogMonitor) BuildSet() {
```
mkeesey, line 131: I think we should do this in a `NewLogMonitor` method that allocates and initializes at the same time. The lazy initialization in `IsMonitored` will be very racy and not memory safe when there are multiple callers which there will be in collector with our sinks.

jessejlt: It looks like LogMonitor instances are created by TOML unmarshalling and not through NewLogMonitor. Updated to make initialization thread-safe.

PR #960, opened 2025-10-06
Initial ClickHouse integration

jessejlt:
In this initial iteration we utilize the existing CSV embedded in WAL segments for ingestion into ClickHouse, which keeps this PR much smaller than if directly adopted the ClickHouse binary format. In a subsequent PR, we'll adopt ClickHouse's binary format as embedded in our WAL segments, which should complete MVP for this task.


ingestor/clickhouse/uploader.go
```diff
@@ -0,0 +1,655 @@
...
+				for _, col := range schemaDef.Columns {
+					columnSpecs = append(columnSpecs, fmt.Sprintf("%s:%s", col.Name, col.Type))
+				}
+				u.log.Info("prepared ClickHouse schema",
+					slog.String("table", batch.Table),
+					slog.String("schema_id", schemaID),
+					slog.Any("columns", columnSpecs))
+			}
+
+			writer, err = u.conn.PrepareInsert(ctx, batch.Database, batch.Table, schemaDef.Columns)
```
jwilder, line 327: I know you plan to revamp this bit, but you might be able to just import the segment directly like we do with adx size and:

INSERT INTO sometable
FROM INFILE 'data_small.csv'
FORMAT CSV

You'd need to be able to stream the segments via the statement though.

jwilder:
Overall looks good.  We should see if there are ways to pull out some of the common pieces in the adx and clickhouse ingestion paths.  Some of the files look like dups of the original.  It does keep the implementations safely separated the way it is now though so maybe not worthwhile.

mkeesey (approved):

PR #974, opened 2025-10-16
Remove hard-coded ingestion batching policies

jwilder:
To adjust these now, management commands should be used that allows more fine grained polcies.


docs/cookbook.md
```diff
@@ -157,6 +157,92 @@ TableName
   - Without delay: Process 13:00-14:00 data
   - With delay: Process 12:00-13:00 data (shifted back by 10 minutes, then aligned to hour boundary)
 
+### Optimizing Ingestion Latency
+
+There are several layers of batching within ADX-Mon that allow turning of throughput and latency.
```
jessejlt:
```suggestion
There are several layers of batching within ADX-Mon that allow tuning of throughput and latency.
```

jessejlt (changes requested):

PR #977, opened 2025-10-17
Enhance Function status conditions

jessejlt:
## Summary
- add Function condition helpers and tests to api types
- update ingestor reconciliation to surface conditions and clear stale errors
- document the diagnostics and regenerate Function CRDs

## Testing
- go test ./api/v1
- go test ./ingestor/adx
- go test ./ingestor/storage

### DIFF
--- a/ingestor/adx/tasks.go
+++ b/ingestor/adx/tasks.go
@@ -256,9 +256,6 @@ func (t *SyncFunctionsTask) shouldSkipDatabaseMismatchUpdate(fn *v1.Function, av
 	if fn == nil {
 		return false
 	}
-	if fn.Status.ObservedGeneration != fn.GetGeneration() {
-		return false
-	}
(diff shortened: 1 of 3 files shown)


ingestor/adx/tasks.go
```diff
@@ -134,58 +134,100 @@ func (t *SyncFunctionsTask) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to list functions: %w", err)
 	}
 	for _, function := range functions {
-
-		if function.Spec.Database != v1.AllDatabases && function.Spec.Database != t.kustoCli.Database() {
+		availableDB := t.kustoCli.Database()
+		configuredDB := function.Spec.Database
+		if configuredDB != v1.AllDatabases && !strings.EqualFold(configuredDB, availableDB) {
```
mkeesey: Is there a way we can set these conditions (database mismatch, criteria mismatch) once instead of n-times for the number of Kusto dbs we have? I think every worker for each database will repeatedly set these every reconcile period for ones that don't have matching configs.

PR #983, opened 2025-10-28
Improve ManagementCommand status conditions

jessejlt:
## Summary
- add shared criteria/owner condition helpers for CRDs
- enrich ManagementCommand status with criteria and execution conditions
- update ingestor task and tests to validate new condition flow

## Testing
- go test ./api/v1 ./ingestor/adx
- go test ./...

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,13 +35,13 @@ require (
 	github.com/pquerna/ffjson v0.0.0-20190930134022-aa0246cd15f7
 	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/client_model v0.6.2
-	github.com/prometheus/common v0.67.1
+	github.com/prometheus/common v0.67.2
 	github.com/siderolabs/go-kmsg v0.1.4
 	github.com/stretchr/testify v1.11.1
 	github.com/tenebris-tech/tail v1.0.5
 	github.com/testcontainers/testcontainers-go v0.39.0
 	github.com/testcontainers/testcontainers-go/modules/k3s v0.39.0
-	github.com/tinylib/msgp v1.4.0
+	github.com/tinylib/msgp v1.5.0
 	github.com/traefik/yaegi v0.16.1
 	github.com/urfave/cli/v2 v2.27.7
 	github.com/valyala/fastjson v1.6.4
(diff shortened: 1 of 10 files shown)


ingestor/adx/tasks.go
```diff
@@ -353,14 +366,24 @@ func (t *ManagementCommandTask) Run(ctx context.Context) error {
...
+			if updErr := t.store.UpdateStatus(ctx, command, fmt.Errorf("%s", parsed)); updErr != nil {
+				logger.Errorf("Failed to update management command status: %v", updErr)
 			}
+			continue
 		}
 
 		logger.Infof("Successfully executed management command %s.%s", command.Spec.Database, command.Name)
-		if err := t.store.UpdateStatus(ctx, &command, nil); err != nil {
+		message := "Management command executed successfully"
+		if endpoint := strings.TrimSpace(t.kustoCli.Endpoint()); endpoint != "" {
```
mkeesey: I'm not sure about this conditional. I think the `kustoCli` instance will pretty much always have an endpoint set on it if we're using it, and that's not really the criteria to know if we're executing against a specific db or not.

ingestor/adx/tasks.go
```diff
@@ -342,8 +348,15 @@ func (t *ManagementCommandTask) Run(ctx context.Context) error {
 				if logger.IsDebug() {
 					logger.Debugf("Skipping management command %s/%s due to criteriaExpression evaluating to false", command.Namespace, command.Name)
 				}
+				command.SetCriteriaMatchCondition(false, expression, nil, t.ClusterLabels)
+				message := fmt.Sprintf("Management command skipped because criteria expression evaluated to false for cluster labels: %s", v1.FormatClusterLabels(t.ClusterLabels))
+				command.SetExecutionCondition(metav1.ConditionFalse, v1.ReasonCriteriaNotMatched, message)
+				if updErr := t.store.UpdateStatus(ctx, command, fmt.Errorf("%s", message)); updErr != nil {
```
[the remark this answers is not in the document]
jessejlt: While this isn't necessarily an error, it is a halting condition and the status message is valid.

ingestor/adx/tasks.go
jessejlt: `StatementExecutor` does not guarantee that the Endpoint is set, so the check is valid. We could short-circuit the check by hoisting it up the chain; however, the risk is that we leave all ManagementCommand CRDs with their last execution status. So if we have dozens of CRDs in some previous execution state, and we short-circuit the kusto client endpoint check, we leave the CRDs in their previous status condition state, which is misleading. I think the intent with controllers is that each CRD is suppose to reflect its own state.

PR #999, opened 2025-11-11
Add ADX operator README

jessejlt:
## Summary
- add operator-focused ADX README covering configuration and federation behavior

## Testing
- not run (docs only)


operator/README.md
```diff
@@ -0,0 +1,275 @@
...
+- When zero-config provisioning is used, the resulting endpoint is written back into the CRD (`spec.endpoint`) once the cluster becomes reachable.
+
+## ADXCluster specification anatomy
+
+### Core cluster fields
+
+| Field | Required | Notes |
+| --- | --- | --- |
+| `spec.clusterName` | ✅ | Azure resource name for the ADX cluster. Must match `^[a-zA-Z0-9-]+$` and be ≤100 characters. |
+| `spec.endpoint` | ⛔ (optional) | Provide the HTTPS URI of an existing cluster to skip provisioning. |
```
mkeesey: The red stop emojis are not the clearest - here it's intending to communicate that it's not required but usually red icons indicate to me that it is a deprecated field or something that should be avoided.  Inconsistent usage of `(optional)` as well. Maybe use the words Required and Optional?

operator/README.md
```diff
@@ -0,0 +1,275 @@
+# ADX Operator Guide
+
+> Scope: this document focuses exclusively on what the adx-mon operator does for Azure Data Explorer (ADX). Ingestor and Collector component operations are still a work-in-progress.
+
+## What the operator provides (ADX only)
```
mkeesey: I think we should have prose describing the problem we solve with federation and a mermaid diagram with example deployments (one or two hubs, with several downstream partition clusters). Some description as well about the data residing in the partitions, but "one pane of glass" to query through the hubs. What problems does the operator solve?

operator/README.md
```diff
@@ -0,0 +1,275 @@
...
+  clusterName: demo-hub
+  endpoint: "https://demo-hub.kusto.windows.net"
+  role: Federated
+  federation:
+    heartbeatDatabase: FleetDiscovery
+    heartbeatTable: Heartbeats
+    heartbeatTTL: 2h
+```
+
+Apply it with `kubectl apply -f hub.yaml`. Because an `endpoint` is provided the operator assumes the cluster already exists; it immediately marks the CRD ready and waits for federation loops to begin. Ensure the hub already has the listed databases or grant the operator permission and provisioning details so later federation passes can create missing assets.
```
mkeesey:
```suggestion
Manages an existing ADX cluster `endpoint`. The operator MSI must have `AllDatabasesAdmin` to create the federation database used for partition heartbeats, or the databases and tables must already exist. This creates functions in the ADX cluster that queries the remote partition clusters for their data.
```

operator/README.md
```diff
@@ -0,0 +1,275 @@
+# ADX Operator Guide
```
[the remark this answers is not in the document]
jessejlt: Ya I can move this into the primary docs. I put it in a readme because the feature isn't yet ready for primetime, but I'll move it

mkeesey (approved):

PR #1005, opened 2025-11-13
jesse/blocked clusters

jessejlt:
- **Improve blocked cluster federation handling**
- **Regenerate ADXCluster CRD**


operator/adx.go
```diff
@@ -1621,16 +1636,190 @@ func splitKustoScripts(funcs []string, maxSize int) [][]string {
...
-		_, err := client.Mgmt(ctx, database, kql.New("").AddUnsafe(fullScript))
-		if err != nil {
-			return fmt.Errorf("failed to execute Kusto script: %w", err)
+		if _, err := client.Mgmt(ctx, database, kql.New("").AddUnsafe(fullScript)); err != nil {
+			logger.Errorf("ADXCluster: failed to execute Kusto script chunk %d/%d for database %s: %v", idx+1, len(scripts), database, err)
+			errs = append(errs, fmt.Errorf("chunk %d: %w", idx+1, err))
 		}
 	}
+	if len(errs) > 0 {
+		return fmt.Errorf("one or more Kusto scripts failed: %w", errors.Join(errs...))
```
mkeesey: This will be an enormous error message if even a couple fail because it's keeping the full outputs and concatenating them all together. Maybe we can filter the messages like we do for updating CRD statuses or just keep track of the title of the function or something. I think throwing this detail away should be fine given that we're logging the full error message up above already

PR #1009, opened 2025-11-14
Create kubelet-discovery config block, fix bugs with kubelet discovery process

mkeesey:
This adds a new top-level configuration property `[kubelet-discovery]` which enables Pod discovery by polling the local node's Kubelet /pods api instead of using Informers against the cluster apiservers. This is intended to greatly improve scalability with larger cluster sizes to reduce the number of persistent watches and connections with many nodes.

This PR also makes the change detection logic much more robust, as the kubelet (at least in v1.33.5) will not always change the `resourceVersion` field for many changes, including container crashes, despite the API server doing so. This necessitates more manual checks for changes to ensure we capture changes to the pods.

This implementation is not currently enabled by default and can have implications around collecting logs from very short-lived pods. If the pod exists and is deleted in a very short period of time between polls of the kubelet api, `host-log` may not discover that pod's metadata and not be able to ingest the logs. This case can also occur with informers, but the polling nature of this change may make it more likely in certain configurations. A more reliable mechanism for this use case would be writing pod logs directly with a logger to the otel endpoint of the collector.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -10,7 +10,7 @@ This is the top level configuration for the collector. The only required fields 
 ```toml
 # Ingestor URL to send collected telemetry.
 endpoint = 'https://ingestor.adx-mon.svc.cluster.local'
-# Path to kubernetes client config
+# Path to kubernetes client config file. If not set, in-cluster config is used for node metadata and pod discovery.
 kube-config = '.kube/config'
 # Skip TLS verification.
 insecure-skip-verify = true


cmd/collector/config/config.go
```diff
@@ -154,6 +155,31 @@ func (s *PrometheusScrape) Validate() error {
...
+		return errors.New("kubelet-discovery.port must be between 0 and 65535")
+	}
+
+	if k.PollInterval != 0 && k.PollInterval < 1 {
+		return errors.New("kubelet-discovery.poll-interval must be at least 1 second")
+	}
+
+	if k.RequestTimeout != 0 && k.RequestTimeout < 1 {
+		return errors.New("kubelet-discovery.request-timeout must be at least 1 second")
+	}
```
jessejlt:
```suggestion
	if k.RequestTimeout != 0 && k.RequestTimeout < 1 {
		return errors.New("kubelet-discovery.request-timeout must be at least 1 second")
	}
	if k.PollInterval <= k.RequestTimeout {
		return errors.New("kubelet-discover.poll-interval must be less than kubelet-discovery.request-timeout")
	}
```

Should we ensure the `PollInterval < RequestTimeout`?

PR #1051, opened 2026-01-06
operator: fix ingestor reconciliation and CRD handling

jessejlt:
## Summary
- add TLS and applied provisioning state fixes for the Ingestor CRD
- refactor ingestor resource rendering to programmatic builders, including pull secrets/node selector/tolerations propagation
- recover from criteria-expression error states across operator controllers
- update ingestor CRD docs

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/golang/snappy v1.0.0
 	github.com/google/cel-go v0.26.1
 	github.com/google/uuid v1.6.0
-	github.com/klauspost/compress v1.18.2
+	github.com/klauspost/compress v1.18.3
 	github.com/klauspost/pgzip v1.2.6
 	github.com/libp2p/go-buffer-pool v0.1.0
 	github.com/lmittmann/tint v1.1.2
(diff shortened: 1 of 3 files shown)


operator/manifests/ingestor_rbac.yaml
```diff
@@ -0,0 +1,65 @@
...
+  labels:
+    app.kubernetes.io/name: {{ .Name }}
+    app.kubernetes.io/component: ingestor
+    app.kubernetes.io/managed-by: adx-mon-operator
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - namespaces
+      - pods
```
mkeesey: nit for future because we've had it like this forever - I think we should move the pods permissions to `RoleBinding` for just our namespace in the future. IIRC ingestor only needs to see its peer pods

tools/crdgen/Dockerfile
```diff
@@ -11,11 +11,15 @@ RUN tdnf install -y ca-certificates && update-ca-trust
 # compiler + headers
 RUN tdnf install -y \
   gcc make binutils glibc-devel kernel-headers pkg-config \
-  systemd-devel \
+  systemd-devel tar \
  && tdnf clean all
 
-RUN tdnf install -y golang
-
+# Install Go (specific version to match go.mod requirements)
```
mkeesey: Surprised we need this - would expect that our other builds would be broken too because we use tdnf install -y golang in those too. I think Go will download the appropriate toolchain if necessary.

operator/ingestor.go
```diff
@@ -388,32 +636,221 @@ func (r *IngestorReconciler) CreateIngestor(ctx context.Context, ingestor *adxmo
...
+		if existing.Spec.Template.Spec.Containers[0].Image != desired.Spec.Template.Spec.Containers[0].Image {
+			logger.Debugf("StatefulSet image differs: existing=%s, desired=%s",
+				existing.Spec.Template.Spec.Containers[0].Image, desired.Spec.Template.Spec.Containers[0].Image)
+			return true
 		}
 	}
 
-	if err := r.setCondition(ctx, ingestor, r.waitForReadyReason, "Ingestor manifests installing", metav1.ConditionTrue); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to set status condition: %w", err)
+	// Compare container args (sorted for comparison)
```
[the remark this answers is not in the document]
jessejlt: All sorting was removed in favor of a set-based comparison of these arguments

PR #1148, opened 2026-05-04
fix(wal): segmentIterator gracefully handles corrupt blocks

mkeesey:
WAL machinery within Collector and Ingestor contains a Repair() mechanism that is called on startup. This mechanism iterates through blocks within a given segment and truncates any files that contain corruption. Once a given block is considered corrupted, it and all of the following blocks are all considered corrupt as well.

If corruption happens during runtime, perhaps from a collector instance that crashed, the segmentIterator.Next() implementation would handle some types of corruption (zero block lengths) gracefully but would return errors for other cases like short blocks or those that don't contain a valid checksum. Instead of returning errors, do a virtual truncation where we report there are no more blocks to iterate to.

### DIFF
--- a/ingestor/adx/uploader_test.go
+++ b/ingestor/adx/uploader_test.go
@@ -3,12 +3,16 @@ package adx
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"errors"
+	"hash/crc32"
 	"io"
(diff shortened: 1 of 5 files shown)


jwilder (approved):
This makes sense.

PR #1175, opened 2026-06-15
perf(ingestor): optimize prom_rate to a single per-series pass

jessejlt:
This is a follow up to a recent meeting regarding the performance of `prom_rate`. The git commit explains the issue with the current implementation and how the optimized version helps. The optimization reduces memory consumption by 30%.


jwilder (approved):
Would a similar chanage help prom_delta and prom_increase?

PR #1193, opened 2026-07-16
refactor(operator): migrate operator to azkustodata

aaqib-m:
Migrates the adxoperator from legacy kusto client to azkustodata, modernizing the operator’s ADX data-plane integration with safer query construction, cleaner result handling, and more efficient client reuse.

### DIFF
--- a/operator/adx.go
+++ b/operator/adx.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
 	"sort"
 	"strings"
@@ -16,8 +15,8 @@ import (
 	"github.com/Azure/adx-mon/pkg/celutil"
 	"github.com/Azure/adx-mon/pkg/kustoutil"
 	"github.com/Azure/adx-mon/pkg/logger"
-	"github.com/Azure/azure-kusto-go/kusto"
-	"github.com/Azure/azure-kusto-go/kusto/kql"
+	"github.com/Azure/azure-kusto-go/azkustodata"
+	"github.com/Azure/azure-kusto-go/azkustodata/kql"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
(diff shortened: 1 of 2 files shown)


operator/adx.go
```diff
@@ -1770,40 +1751,33 @@ func ensureHubTables(ctx context.Context, client *kusto.Client, database string,
...
+		if !tbl.IsPrimaryResult() {
 			continue
 		}
-		if errFinal != nil {
-			return false, fmt.Errorf("failed to read table existence: %w", errFinal)
-		}
-		var rec DatabaseExistsRec
-		if err := row.ToStruct(&rec); err != nil {
-			return false, fmt.Errorf("failed to parse table existence: %w", err)
+		for _, row := range tbl.Rows() {
```
mkeesey: For this path, checkIfFunctionIsView, and executeKustoScripts (and maybe others) where we only expect one row and return within the `for _, row...` loop, I think it'll be clearer to avoid the loop. Just assert that the length of `tbl.Rows()` == 1, then extract the single value and return.

operator/adx_test.go
```diff
@@ -80,6 +80,17 @@ func TestWrapHeartbeatStageError(t *testing.T) {
 	}
 }
 
+func TestBuildHeartbeatQuery(t *testing.T) {
+	t.Parallel()
+
+	query, err := buildHeartbeatQuery("Heartbeat Table", "1h30m")
```
mkeesey: nit - let's use a valid identifier for tables. I don't think spaces are allowed - perhaps `HeartbeatTable`

operator/adx.go
aaqib-m: Updated these paths to require exactly one primary result row and error on missing or malformed results. Decided to pull the shared check into a small helper since it's duplicated a few times

operator/adx.go
```diff
@@ -1252,31 +1239,24 @@ func listHeartbeatFunctions(ctx context.Context, client *kusto.Client, database
...
-
-		var fn FunctionRec
-		if err := row.ToStruct(&fn); err != nil {
-			return nil, wrapHeartbeatStageError(stage, heartbeatListFunctionsTimeout, fmt.Errorf("parse function: %w", err))
+		for _, row := range table.Rows() {
+			var fn FunctionRec
+			if err := row.ToStruct(&fn); err != nil {
+				return nil, wrapHeartbeatStageError(stage, heartbeatListFunctionsTimeout, fmt.Errorf("parse function: %w", err))
+			}
+			functions = append(functions, fn)
```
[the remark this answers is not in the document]
aaqib-m: good suggestion; changed the query to only fetch the name

mkeesey (approved):

PR #1194, opened 2026-07-16
refactor(adxexporter): migrate adxexporter to azkustodata

aaqib-m:
Migrated adxexporter from legacy Kusto client to azkustodata, modernizing the exporter’s ADX integration with safer query handling, cleaner result parsing, and a simpler azkustodata-based client interface.

### DIFF
--- a/adxexporter/kusto.go
+++ b/adxexporter/kusto.go
@@ -3,6 +3,7 @@ package adxexporter
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
@@ -10,6 +11,7 @@ import (
 	azkustodata "github.com/Azure/azure-kusto-go/azkustodata"
 	"github.com/Azure/azure-kusto-go/azkustodata/kql"
 	azquery "github.com/Azure/azure-kusto-go/azkustodata/query"
+	azvalue "github.com/Azure/azure-kusto-go/azkustodata/value"
 	"k8s.io/utils/clock"
 )
 
@@ -20,8 +22,8 @@ type KustoExecutor interface {
 	Database() string
 	// Endpoint returns the Kusto cluster endpoint
 	Endpoint() string
-	// Query executes a KQL query and returns the results
-	Query(ctx context.Context, query azkustodata.Statement, options ...azkustodata.QueryOption) (azquery.Dataset, error)
+	// IterativeQuery executes a KQL query and streams the results
+	IterativeQuery(ctx context.Context, query azkustodata.Statement, options ...azkustodata.QueryOption) (azquery.IterativeDataset, error)
 	// Mgmt executes a Kusto management command (dot-command)
 	Mgmt(ctx context.Context, query azkustodata.Statement, options ...azkustodata.QueryOption) (azquery.Dataset, error)
 }
@@ -63,8 +65,8 @@ func (k *KustoClient) Endpoint() string {
 	return k.endpoint
 }
 
-func (k *KustoClient) Query(ctx context.Context, query azkustodata.Statement, options ...azkustodata.QueryOption) (azquery.Dataset, error) {
-	return k.client.Query(ctx, k.database, query, options...)
+func (k *KustoClient) IterativeQuery(ctx context.Context, query azkustodata.Statement, options ...azkustodata.QueryOption) (azquery.IterativeDataset, error) {
+	return k.client.IterativeQuery(ctx, k.database, query, options...)
 }
 
 // Mgmt executes a Kusto management command (dot-command) against the configured database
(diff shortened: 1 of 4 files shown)


adxexporter/kusto.go
```diff
@@ -139,37 +138,35 @@ func (qe *QueryExecutor) ExecuteQuery(ctx context.Context, queryBody string, sta
...
-				rowMap[colName] = row.Values[i]
+			// Convert row to map.
+			rowMap := make(map[string]interface{})
+			columns := row.Columns()
+			values := row.Values()
+			for i, col := range columns {
+				if i < len(values) {
+					rowMap[col.Name()] = values[i]
+				}
 			}
```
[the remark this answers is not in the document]
mkeesey: I think our current tests are probably insufficient to catch this - we should definitely add tests to make sure downstream stuff gets expected types, etc.

If tests for `ExecuteQuery` aren't able to catch this kind of thing, we should get some added.

adxexporter/kusto.go
```diff
@@ -127,10 +127,9 @@ func (qe *QueryExecutor) ExecuteQuery(ctx context.Context, queryBody string, sta
 			Duration: qe.clock.Since(start),
 		}, nil
 	}
-	defer iter.Stop()
 
 	// Convert results to rows
-	rows, err := qe.iteratorToRows(iter)
+	rows, err := qe.datasetToRows(iter)
```
mkeesey: In this conversion I think we should use the iterative query path that the Kusto client supports instead of `azquery.Dataset`. This should be supported because we aren't doing a `mgmt` query.

Doing it iteratively means we don't read everything into memory all at once as we iterate through which can be a huge amount of data, especially if the result returns too many rows. This should help prevent OOMs due to unfortunate export queries, etc.

adxexporter/summaryrule.go
```diff
@@ -329,24 +328,30 @@ func (r *SummaryRuleReconciler) updateSummaryRuleStatus(rule *adxmonv1.SummaryRu
...
+	for _, table := range ds.Tables() {
+		if !table.IsPrimaryResult() {
 			continue
 		}
-		if errFinal != nil {
-			return "", fmt.Errorf("failed to retrieve operation ID: %v", errFinal)
-		}
-		if len(row.Values) != 1 {
-			return "", fmt.Errorf("unexpected number of values in row: %d", len(row.Values))
+		for _, row := range table.Rows() {
```
mkeesey: Instead of looping through rows when we expect one, assert `table.Rows()` has size 1 like we do for values below.

adxexporter/summaryrule.go
```diff
@@ -329,24 +328,30 @@ func (r *SummaryRuleReconciler) updateSummaryRuleStatus(rule *adxmonv1.SummaryRu
...
+			if len(values) > 1 {
+				return "", fmt.Errorf("unexpected number of values in row without OperationId column: %d", len(values))
+			}
+
+			return values[0].String(), nil
 		}
-		return row.Values[0].String(), nil
+		break
 	}
 	return "", nil
```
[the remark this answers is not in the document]
mkeesey: agreed - old bug but important one I think.

mkeesey (changes requested):

adxexporter/kusto.go
```diff
@@ -129,7 +131,10 @@ func (qe *QueryExecutor) ExecuteQuery(ctx context.Context, queryBody string, sta
 	}
 
 	// Convert results to rows
-	rows, err := qe.datasetToRows(iter)
+	rows, err := qe.iterativeDatasetToRows(dataset)
```
mkeesey, line 135: We lose the error message from `err`, and if we close successfully but have an error from the query results, I think we still return an (invalid) QueryResult.

Think we probably need some kind of test for this path too.

I don't think we care as much about errors from Close(), so the code can look like:
```
//after the err check above for InterativeQuery
defer dataset.Close()
rows, err := qe.iterativeDatasetToRows(dataset)
if err != nil {
  return nil, err
}
...
```

adxexporter/kusto.go
```diff
@@ -138,36 +143,78 @@ func (qe *QueryExecutor) ExecuteQuery(ctx context.Context, queryBody string, sta
...
-					rowMap[col.Name()] = values[i]
+					rowMap[col.Name()] = materializeKustoValue(values[i])
 				}
 			}
 			rows = append(rows, rowMap)
 		}
 
-		// Only process the primary result table; exclude metadata tables.
-		break
+		if primary {
```
mkeesey: I think this primary check is unnecessary - we check for primary higher in the loop and skip the body if it isn't.

mkeesey (changes requested):

mkeesey (approved):

PR #1195, opened 2026-07-19
refactor(transform): remove legacy Kusto value wrappers

aaqib-m:
Removes the transformer’s dependency on legacy wrappers and updates tests to use the native Go values produced by adxexporter.

The transformer intentionally does not accept `azkustodata` wrappers. SDK materialization is owned by adxexporter, keeping the transformer SDK-independent and ensuring wrapper leaks fail fast instead of duplicating conversion and null-handling logic.

Null behavior remains explicit through native nil, and existing support for numeric values, strings, timestamps, and labels is preserved.

**Depends on the ADX exporter migration being merged first.**

### DIFF
--- a/transform/kusto_to_metrics.go
+++ b/transform/kusto_to_metrics.go
@@ -403,7 +403,7 @@ func (t *KustoToMetricsTransformer) extractLabels(row map[string]any) (map[strin
 		case string:
 			labels[labelColumn] = v
 		default:
-			// Lables must be string key:value pairs.
+			// Labels must be string key:value pairs.
 			return nil, fmt.Errorf("label column '%s' contains unsupported type %T", labelColumn, rawValue)
 		}
 	}
--- a/transform/kusto_to_metrics.go
+++ b/transform/kusto_to_metrics.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/Azure/adx-mon/pkg/logger"
 	"github.com/Azure/adx-mon/pkg/prompb"
-	"github.com/Azure/azure-kusto-go/kusto/data/value"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 )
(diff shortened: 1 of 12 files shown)


transform/kusto_to_metrics.go
```diff
@@ -178,8 +177,8 @@ func constructMetricName(baseName, prefix, columnName string) (string, error) {
 	return result, nil
 }
 
-// Transform converts KQL query results to Prometheus metrics
-// Supports both single-value and multi-value column modes
+// Transform converts materialized KQL rows to Prometheus metrics.
+// Callers must convert Kusto SDK values and nulls to native Go values first.
```
mkeesey, line 181: This is a new requirement that doesn't appear to be honored by callers - I think this is called directly from the results of calls to `iteratorToRows`. I think this probably will fail during runtime because we're passing in Kusto values and downstream stuff no longer gets primitives like they expect.

The typing of this method is pretty rough which helps mask this - I think the parameter should be something like `results []map[string]value.Kusto` to make this much more strongly typed to emphasize exactly what kind of data we're passing in.

The existing impl is a mess (it assumes primitives and Kusto values) - I think it's important we clean this up to make it more robust.

We should also add some kind of work item to add an e2e test for this kind of flow as well.

transform/kusto_to_metrics_test.go
```diff
@@ -1608,21 +1577,6 @@ func TestExtractValues(t *testing.T) {
 			},
 			wantErr: false,
 		},
-		{
-			name: "extraction with Kusto value types",
-			row: map[string]any{
-				"LongValue": value.Long{Value: 12345, Valid: true},
```
mkeesey, line 1614: These are the ones that definitely shouldn't be removed I don't think.

mkeesey (changes requested):

mkeesey (approved):

PR #1196, opened 2026-07-20
Add reusable alerter alert primitives

mkeesey:
## Summary
- allow query contexts to use explicit start and end times
- extract alert query-row parsing into a reusable, delivery-neutral helper
- validate malformed rows and preserve existing alert delivery behavior

## Testing
- `go test ./alerter/engine`

### DIFF
--- a/alerter/engine/context.go
+++ b/alerter/engine/context.go
@@ -26,10 +26,17 @@ type queryParam struct {
 }
 
 func NewQueryContext(rule *rules.Rule, endTime time.Time, region string) (*QueryContext, error) {
+	if rule == nil {
+		return nil, fmt.Errorf("rule must not be nil")
+	}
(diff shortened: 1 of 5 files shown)


alerter/engine/context.go
```diff
@@ -26,10 +26,14 @@ type queryParam struct {
 }
 
 func NewQueryContext(rule *rules.Rule, endTime time.Time, region string) (*QueryContext, error) {
+	return NewQueryContextForWindow(rule, endTime.Add(-rule.Interval), endTime, region)
+}
+
+func NewQueryContextForWindow(rule *rules.Rule, startTime, endTime time.Time, region string) (*QueryContext, error) {
```
aaqib-m, line 35: nit: might be worthwhile to add a nil check for the rule pointer

aaqib-m (approved):

PR #1199, opened 2026-07-21
feat(alerter): add AlertRule performance metrics

aaqib-m:
Adds per-AlertRule evaluation duration, outcome, and generated-alert metrics alongside a structured evaluation summary.


alerter/engine/worker.go
```diff
@@ -186,15 +192,42 @@ func (e *worker) ExecuteQuery(ctx context.Context) {
 	ctx, cancel := context.WithTimeout(ctx, maxQueryTime)
 	defer cancel()
 
-	endTime := time.Now().UTC()
+	executionTime := time.Now().UTC()
+	evaluationStart := time.Now()
```
mkeesey: This contains a lot of various bits of state with many separate variables and some relatively big anonymous function blocks, all for capturing evaluation times, etc. These aren't as relevant to the specific bits of what this code is doing within the worker, so I think we should encapsulate it into a new struct/object that we use sort of like this (the name is bad - you can probably come up with a better one 😄)

```
evaluation := NewEvaluation() //sets up the initial time, etc
defer evaluation.Finish() // does the telemetry sending, perhaps the alertrule status update or we pass this to the alertruleStatusUpdate
...

evaluation.Outcome = evalulationOutcomeSetupError // update these variables given the various states, consumed by Finish
evaluation.Rows = 0
...
```

aaqib-m: good point! Didn't really have much of a better name, maybe slightly longer if anything 😅

mkeesey (approved):

PR #1205, opened 2026-08-05
changes after adx-collector pprof  (in prod) for CPU optimization

Elaa-jamazi:
**Summary**

Optimizes CPU/allocation overhead in Collector, driven by pprof profiling of collector pods in production, which identified excessive per-request allocations in protobuf decode and remote-write handling as a top CPU contributor.

**Changes**

•  pkg/prompb/protobuf.go ,  pkg/prompb/protobuf_test.go : Reuse  Label / Sample / TimeSeries  objects during protobuf decode instead of allocating fresh ones per request; return  WriteRequest s and dropped series to pools after use.
•  collector/metrics/handler.go ,  collector/metrics/handler_test.go : Wire pooled objects back through the write path so pooled requests are safely reused after a request completes.
•  collector/scraper.go ,  collector/otlp/metrics.go : Adjust the scraper and OTLP ingestion paths to participate correctly in the pooling lifecycle (no double-pooling/use-after-reset).
•  transform/transformer.go ,  transform/transformer_test.go : Preserve reusable label objects when transformations drop labels or series, avoiding unnecessary reallocation.

### DIFF
--- a/collector/metrics/handler.go
+++ b/collector/metrics/handler.go
@@ -127,8 +127,13 @@ func (s *Handler) HandleReceive(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Note: this cause allocations, but holding onto them in a pool causes a lot of memory to be used over time.
-	req := prompb.WriteRequestPool.Get()
+	// Note: this causes an allocation, but holding onto WriteRequests in a pool causes a lot of
+	// memory to be used over time, since senders can flush very large backlogs (e.g. after a
+	// restart) and there's no bound on how many outsized requests the pool could retain at once.
+	// The individual TimeSeries/Label/Sample objects decoded below are still pooled and bounded
+	// via TimeSeriesPool, which is where the allocation savings actually come from.
+	req := &prompb.WriteRequest{}
+	defer req.Reset()
 	if err := req.Unmarshal(reqBuf); err != nil {
 		m.WithLabelValues(strconv.Itoa(http.StatusBadRequest)).Inc()
 		http.Error(w, err.Error(), http.StatusBadRequest)
--- a/storage/store.go
+++ b/storage/store.go
@@ -40,8 +40,6 @@ var (
 	})
 )
 
-const maxNativeLogsBatchSize = 16 * 1024
-
 type Store interface {
 	service.Component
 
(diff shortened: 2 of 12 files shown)


collector/metrics/handler.go
```diff
@@ -127,8 +127,10 @@ func (s *Handler) HandleReceive(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Note: this cause allocations, but holding onto them in a pool causes a lot of memory to be used over time.
 	req := prompb.WriteRequestPool.Get()
+	defer func() {
```
mkeesey: This change unfortunately ignores the comment above that was removed. We should probably create the write requests directly instead of the pool so it's not as confusing, but this can be problematic. WriteRequests contain collections of objects which can contain many many objects, especially after restarts when senders flush a lot of data.

Retaining WriteRequests that have accumulated a lot of data can cause a collector to easily exceed its heap even when we don't need all of that space for subsequent operations later. We'll still be holding onto them within the pool.

storage/store.go
```diff
@@ -207,21 +209,45 @@ func (s *LocalStore) WriteOTLPLogs(ctx context.Context, database, table string,
 
 func (s *LocalStore) WriteNativeLogs(ctx context.Context, logs *types.LogBatch) error {
 	enc := nativeLogsCSVWriterPool.Get(8 * 1024).(*transform2.NativeLogsCSVWriter)
-	defer nativeLogsCSVWriterPool.Put(enc)
+	defer func() {
+		enc.Reset()
```
mkeesey: This appears to add several different encoder resets with different lifecycles than we typically do. It also changes quite a lot of the logic in ways that I'm not very sure is safe. This store implementation is pretty heavily tested and benchmarked and deserves a lot of scrutiny.

collector/metrics/handler.go
Elaa-jamazi: I addressed this issue. Now  requests write request like you recommended: req := &prompb.WriteRequest{}  +  defer req.Reset()  — allocates the outer struct directly (cheap; just slice headers) and still releases its inner  TimeSeries / Label / Sample  objects back to the bounded  TimeSeriesPool  via  Reset()

storage/store.go
Elaa-jamazi: As we discussed on teams, I'll be removing the changes made to the log path, for now

matucker-msft:
@Elaa-jamazi  please sign the CLA so we can review this change. Thanks.

collector/metrics/handler.go
Elaa-jamazi: what if we add a boundedpool to avoid a pool causing a lot of memory to be used over time and to avoid the collection of objects concern with the WriteRequests? Each pooled  TimeSeries  is now capped at ~516 KiB (64 labels × 8 KiB max each, plus ~4 KiB of samples, if a Timeseries is oversized  it's left for normal Go GC instead ). A new  BoundedPool  caps the pool to at most 1024 retained items via a fixed-capacity channel — anything beyond that is dropped instead of pooled. Multiplying these two hard limits (1024 × 516 KiB ≈ 530 MiB) gives a true worst-case ceiling, not an average, so the pool's memory footprint can never exceed that regardless of traffic burst size.

PR #1207, opened 2026-08-13
Harden ingestor Kubernetes permissions

jwilder:
## Summary
- replace cluster-wide pod and namespace writes with namespace-local pod observation and patch access
- use short-lived projected service account tokens instead of automatic token mounting
- migrate existing operator-managed installations fail-closed, reconcile RBAC drift, and clean up managed cluster RBAC
- tighten CRD permissions and preserve graceful-shutdown API compatibility

## Upgrade notes
- production operator RBAC is maintained in a separate deployment repository and must grant the permissions documented in `docs/designs/operator.md` before this operator version is deployed
- Kubernetes 1.20 or newer is required
- StatefulSets using `OnDelete` require existing pods to be manually replaced
- namespace-wide pod `patch` remains an accepted residual risk for graceful-shutdown completion annotations

## Validation
- `go test ./cmd/operator ./cmd/ingestor ./ingestor/... ./operator`
- `go vet ./operator ./ingestor/runner/shutdown`
- `git diff --check`
- verified generated bundle contains the updated ingestor manifest


build/k8s/ingestor.yaml
```diff
@@ -185,6 +212,24 @@ spec:
                     values:
                       - aks-system
       volumes:
+        - name: kube-api-access
```
matucker-msft: On Kubernetes 1.22+, automatic ServiceAccount credential injection already uses a rotating, Pod-bound projected token with a roughly one-hour lifetime. Since the ingestor Pod has only one container, this explicit projection appears equivalent in lifetime, audience, exposure, and mount location. https://kubernetes.io/docs/concepts/security/service-accounts/#assign-to-pod

operator/ingestor.go
```diff
@@ -46,6 +56,20 @@ func (r *IngestorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	if !ingestor.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, r.cleanupIngestorSecurity(ctx, ingestor)
+	}
+
+	if err := r.reconcileIngestorSecurity(ctx, ingestor); err != nil {
+		_ = r.setCondition(ctx, ingestor, "SecurityMigrationFailed", err.Error(), metav1.ConditionFalse)
+		return ctrl.Result{}, fmt.Errorf("failed to reconcile ingestor security controls: %w", err)
+	}
+	if condition := meta.FindStatusCondition(ingestor.Status.Conditions, adxmonv1.IngestorConditionOwner); condition != nil && condition.Reason == "SecurityMigrationFailed" {
+		if err := r.setCondition(ctx, ingestor, "SecurityMigrationComplete", "Ingestor security controls are reconciled", metav1.ConditionTrue); err != nil {
```
matucker-msft: Could security migration use a separate condition type here instead of IngestorConditionOwner?

setCondition replaces the existing owner condition and sets ObservedGeneration to the current generation. For example, if a spec change creates generation 8 and security reconciliation fails before reaching the lifecycle switch below, SecurityMigrationFailed records generation 8. On retry, this call replaces it with SecurityMigrationComplete=True, also at generation 8. The switch then sees neither WaitForReady, Unknown, nor an observed-generation mismatch, so generation 8’s manifests are never reconciled. A transient failure during installation can similarly replace WaitForReady and prevent further IsReady checks.

Could we track this with a separate condition type, such as IngestorSecurityReady, so the normal lifecycle condition and its observed generation remain intact?

matucker-msft (changes requested):

build/k8s/ingestor.yaml
jwilder: Agreed. I removed the explicit projection and StatefulSet token migration. The pod now opts into Kubernetes 1.22+ native rotating, Pod-bound credential injection while the ServiceAccount default remains false. This keeps token exposure limited to the ingestor workload without duplicating kubelet behavior.

operator/ingestor.go
jwilder: Fixed in 437521eb. Security reconciliation now writes a separate `SecurityReady` condition, leaving `IngestorConditionOwner` and its observed generation untouched. Added regression coverage for both a pending generation and `WaitForReady` across failure and recovery.

jwilder:
Addressed the requested changes in `437521eb`:

- removed the redundant explicit token projection and now rely on Kubernetes 1.22+ native rotating Pod-bound credentials
- moved security migration status to an independent `SecurityReady` condition
- added regression tests proving pending generations and `WaitForReady` lifecycle state survive security failure and recovery

Validation rerun:
- `go test ./cmd/operator ./cmd/ingestor ./ingestor/... ./operator`
- `go vet ./operator ./ingestor/runner/shutdown`
- `git diff --check`
- generated bundle verified

build/k8s/ingestor.yaml
```diff
@@ -12,31 +12,58 @@ metadata:
...
+  - kind: ServiceAccount
+    name: ingestor
+    namespace: adx-mon
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: adx-mon:ingestor
+  labels:
+    adx-mon.azure.com/managed-by: ingestor-operator
```
matucker-msft: I feel like adding `managed-by: ingestor-operator` to the files in the bundle is misleading as it's not managed by the operator. People using the bundle arent installing the operator, they are just installing the raw manifests.

matucker-msft (approved):

build/k8s/ingestor.yaml
jwilder: Agreed. Removed the `adx-mon.azure.com/managed-by: ingestor-operator` labels from the standalone `build/k8s/ingestor.yaml` ClusterRole and ClusterRoleBinding in ae55c8a4, and regenerated `bundle.sh`. The labels remain only in the operator-rendered manifest where the operator actually owns and reconciles those resources.

matucker-msft (approved):

PR #1236, opened 2026-08-26
feat(ingestor): consolidate and aggregate function reconciliation

aaqib-m:
Function reconciliation previously ran independently for each configured Kusto database. Wildcard Functions could therefore be processed by multiple scheduler tasks, causing competing Kubernetes status updates, stale `resourceVersion` conflicts, and noisy logs.

This change consolidates Function processing into one reconciler that executes Kusto operations across matching targets concurrently, aggregates their results, and performs one status update.

- Consolidate Function reconciliation into one scheduler task across all Kusto targets
- Execute target updates concurrently with bounded concurrency and timeouts
- Aggregate target outcomes into one Kubernetes status update
- Retry status and finalizer updates on Kubernetes conflicts using fresh resources
- Propagate scheduler cancellation and wait for active tasks during shutdown


ingestor/adx/tasks.go
```diff
@@ -223,45 +218,152 @@ func (t *SyncFunctionsTask) Run(ctx context.Context) error {
...
+func reconcileFunctionTargets(ctx context.Context, function *v1.Function, targets []StatementExecutor) functionReconcileResult {
+	results := make([]functionTargetResult, len(targets))
+	var group errgroup.Group
+	group.SetLimit(maxConcurrentFunctionTargets)
+	for i, target := range targets {
+		group.Go(func() error {
+			results[i] = executeFunctionTarget(ctx, function.Spec.Body, target)
+			return nil
+		})
+	}
```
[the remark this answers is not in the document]
mkeesey: Not true since Go 1.22.

ingestor/adx/tasks.go
```diff
@@ -223,45 +218,152 @@ func (t *SyncFunctionsTask) Run(ctx context.Context) error {
...
+		return functionReconcileResult{
+			status:          v1.Failed,
+			conditionStatus: metav1.ConditionFalse,
+			reason:          "KustoExecutionRetrying",
+			message:         message,
+			err:             stderrors.New(message),
+		}
+	}
+	if permanentFailures > 0 {
+		message := fmt.Sprintf("Function permanently failed on %d of %d targets", permanentFailures, len(results))
```
mkeesey: One change this PR makes is it removes the specific error message we parsed out of the Kusto response from the status line and changed to this more generic form.

We do have character limits and such, but would it be possible to take say the first one and truncate it (the old code truncated at 200 chars, so we can do something similar) so we can preserve some of the error here?

ingestor/adx/tasks.go
```diff
@@ -193,6 +190,7 @@ func (t *SyncFunctionsTask) Run(ctx context.Context) error {
 				message := fmt.Sprintf("Function skipped because criteria expression evaluated to false for cluster labels: %s", v1.FormatClusterLabels(t.ClusterLabels))
 				function.SetReconcileCondition(metav1.ConditionFalse, "CriteriaNotMatched", message)
 				function.Status.Status = v1.Failed
+				function.Status.ObservedGeneration = function.GetGeneration()
```
mkeesey, line 201: This change will mean that if criteria didn't match before, but we change the server-side criteria (because we goofed or something) we never reevaluate these functions.

I think this change is right, but I think we _also_ should have a mechanism with ListFunctions in store to allow startup reconciliation _once_ where if the reason was `CriteriaNotMatched` we should allow reevaluation.

mkeesey (approved):

PR #1256, opened 2026-09-14
Add common label foundations

mkeesey:
This adds a new CommonLabels field to WriteRequest that is currently not assigned to. This will be used to store a set of static labels that apply to every time series, such as hostnames that collector is configured to apply to every timeseries.

The intention in future PRs is to use this mechanism to reduce allocations for each TimeSeries object. We previously appended all of these common labels to each TimeSeries, but we should preserve one common copy that we simply copy a pointer to in each WriteRequest container.

We have code to lazily iterate through both common and TimeSeries specific labels to support per-series filtering in future PRs without making common labels mutable.

This also fixes some bugs we had in this area including incorrect size calculations (which are not used in prod code anyhow), and ensuring the labels are consistently sorted in all paths which was an existing invarient the metrics code expected.


pkg/prompb/sort.go
```diff
@@ -19,10 +19,7 @@ func (l Labels) Swap(i, j int) {
 
 // IsSorted return true if the labels are sorted according to Sort.
 func IsSorted(l []*Label) bool {
-	if len(l) == 1 {
```
jwilder, line 22: This was an intentional optimization IIRC.  Maybe double check the benchmarks to see if it regresses w/o it.  IsSorted is in the write hot path.

jwilder (approved):

pkg/prompb/merge.go
```diff
@@ -0,0 +1,52 @@
...
+// commonLabels without materializing the merged slice. If both inputs contain
+// the same label name, the per-series label takes precedence.
+//
+// Both input slices must be sorted according to Sort and must not be modified
+// while the iterator is running.
+func MergedLabels(seriesLabels, commonLabels []*Label) iter.Seq[*Label] {
+	return func(yield func(*Label) bool) {
+		series, common := seriesLabels, commonLabels
+		for len(series) > 0 && len(common) > 0 {
+			if bytes.Equal(series[0].Name, common[0].Name) {
```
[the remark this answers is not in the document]
mkeesey: This is expected IMO

pkg/prompb/sort.go
mkeesey: On my Ice Lake machine it makes a pretty small difference, but I'll re-add it. Seems worth it, and not confusing. I'm also changing to be `<= 1` for the empty cases as well - seems to go a bit faster.

aaqib-m (approved):

PR #1257, opened 2026-09-14
feat(alerter): enrich throttled notification summaries

matucker-msft:
## Summary
Enriches throttled notifications with a table of suppressed alert severities and titles, excluding alerts already delivered. Includes up to 100 entries, an omitted-alert count when needed, and Kusto links plus expandable query text for the evaluated time window. Warns when results could not be fully processed.

### Example Notification
**Alert namespace/rule has too many notifications in eastus**

This alert has been throttled by ADX-Mon due to too many notifications. Please reduce the number of notifications for this alert.

Alerts suppressed by throttling:

| Severity | Title |
|---|---|
| 2 | High CPU usage on node-01 |
| 3 | Elevated request latency |

*When truncated:* 5 additional suppressed alerts are not shown. Use the query below to view all results.

Execute in **[Web] [Desktop] [Desktop (SAW)]**

**Click here to show query** expands to display the evaluated query.


alerter/engine/types.go
```diff
@@ -2,8 +2,38 @@ package engine
 
 import (
 	"math"
+
+	"github.com/Azure/adx-mon/alerter/alert"
 )
 
+const maxThrottledNotificationDetails = 100
```
mkeesey, line 9: I think it might make sense to lower this to something like 10-15 given the current limits.

mkeesey (approved):

mkeesey (approved):

